PR #8, opened 2022-12-02
Factor User into a top-level repeatable annotation

tombentley:
The idea being that @User could also be used to support a @SaslScramSha annotation, and maybe @SaslOauthBearer too.

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/User.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/User.java
@@ -6,21 +6,32 @@
 package io.kroxylicious.testing.kafka.common;
 
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import io.kroxylicious.testing.kafka.api.KafkaClusterConstraint;
-import io.kroxylicious.testing.kafka.api.KafkaClusterProvisioningStrategy;
 
-/**
- * Annotation constraining a {@link KafkaClusterProvisioningStrategy} to use
- * provide a cluster that supports SASL-PLAIN. {@link User @User} annotations can be used to
- * configure the cluster with users.
- */
 @Target({ ElementType.PARAMETER, ElementType.FIELD })
 @Retention(RetentionPolicy.RUNTIME)
+@Repeatable(User.List.class)
 @KafkaClusterConstraint
-public @interface SaslPlainAuth {
+public @interface User {
+    /**
+     * @return A user name.
+     */
+    String user();
 
+    /**
+     * @return A password.
+     */
+    String password();
+
+    @Target({ ElementType.PARAMETER, ElementType.FIELD })
+    @Retention(RetentionPolicy.RUNTIME)
+    @KafkaClusterConstraint
+    @interface List {
+        User[] value();
+    }
 }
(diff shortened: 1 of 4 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/common/User.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.kroxylicious.testing.kafka.api.KafkaClusterConstraint;
+
+@Target({ ElementType.PARAMETER, ElementType.FIELD })
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(User.List.class)
+@KafkaClusterConstraint
+public @interface User {
```
SamBarker, line 20: How about `Credentials` rather than `User`?

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
```diff
@@ -141,11 +142,25 @@ public void kraftBasedClusterParameter(@BrokerCluster @KRaftCluster KafkaCluster
...
+        assertInstanceOf(SaslAuthenticationException.class, ee.getCause());
+
+        ee = assertThrows(ExecutionException.class, () -> describeCluster(cluster.getKafkaClientConfiguration("eve", "quux")),
+                "Expect unknown user to throw");
+        assertInstanceOf(SaslAuthenticationException.class, ee.getCause());
+    }
+
+    @Test
+    public void saslPlainAuthenticatingClusterParameterTwoUsers(
+                                                                @BrokerCluster @SaslPlainAuth @User(user = "alice", password = "foo") @User(user = "bob", password = "bar") KafkaCluster cluster)
```
SamBarker, line 163: I found the old version easier to think about. 

```java
@SaslPlainAuth({
                                                                 @SaslPlainAuth.UserPassword(user = "alice", password = "foo"),
                                                                 @SaslPlainAuth.UserPassword(user = "bob", password = "bar")
                                                         }
```

Clearly says to me that these users are authenticated via Sasl Plain. Where as the new version you can assume the same but it's less explicit.

SamBarker (commented):
Given the move to express users independently of the auth mechanism how would one express a test using multiple mechanisms? I'm thinking of cases trying to test that each mechanism is available in parallel?

The simplest way to me would be to express it as:
- `SASL_PLAIN` -> `user1` with `password1`
- `SASL_SCRAM_SHA` -> `user2` with `password2` 

Then assert that each user only works with the appropriate mechanism.

tombentley:
> Given the move to express users independently of the auth mechanism how would one express a test using multiple mechanisms? I'm thinking of cases trying to test that each mechanism is available in parallel?

How much do we care? I mean, there are two ways you could implement that: Via two listeners with a different mechanism for each one, or a single listener with multiple mechanisms. We _could_ allow expressing both of those via annotations. But should we? 

How many test cases are there where that's something you actually care about? I think 90% of the value is being able to execute test cases with a particular mechanism. If you need to test multiple mechanisms then you can use `@TestTemplate` or write two `@Test`s. It's only if you have a single test scenario where you need to authenticate with multiple mechanisms that there's a problem, and I suspect that's a tiny, tiny fraction of the tests people need to write. It might be needed in Apache Kafka itself, but probably vanishingly rare for ecosystem projects.

k-wall:
+1 from me.  I like the separation of concerns - the declaration of of the user/password is separated from the means used to send those over the wire (the SASL mechanism).

impl/src/main/java/io/kroxylicious/testing/kafka/common/User.java
k-wall: User seems okay to me.

Neustradamus:
To follow

k-wall:
I think this is stale. I'm going to close it.

 We have `@SaslMechanism` annotation which defines a set of `@Principal`s.  We don't yet have the ability to sasl mechs to listeners. #367 started to think about that.

```
   void saslAuthWithManyUsers(@BrokerCluster @SaslMechanism(value = "SCRAM-SHA-256", principals = { @Principal(user = "alice", password = "foo"),
            @Principal(user = "bob", password = "bar") }) KafkaCluster cluster) 
...
```

Issue #7, opened 2022-12-02
ParameterExtensionTest is flaky / race condition


tombentley:
Reported by @SamBarker. Example failure:

```
Formatting /var/folders/1g/2nsqtt916vg8f5bftn1tp0580000gn/T/kafka4886149283650868601/broker-0 with metadata.version 3.3-IV0.
Formatting /var/folders/1g/2nsqtt916vg8f5bftn1tp0580000gn/T/kafka7298719589906457285/broker-0 with metadata.version 3.3-IV0.
Formatting /var/folders/1g/2nsqtt916vg8f5bftn1tp0580000gn/T/kafka7298719589906457285/broker-1 with metadata.version 3.3-IV0.

org.opentest4j.AssertionFailedError: 
Expected :2
Actual   :1
<Click to see difference>


	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:528)
	at io.kroxylicious.testing.kafka.junit5ext.ParameterExtensionTest.twoDefinedClusterParameterAndAdmin(ParameterExtensionTest.java:104)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
[...]

PR #9, opened 2022-12-05
apply duct-tape to `InVMKafkaCluster.start` to ensure all nodes are p…

SamBarker:
…resent in the cluster before injecting.

Fixes: #7

### DIFF
--- a/api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
+++ b/api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
@@ -6,7 +6,9 @@
 
 package io.kroxylicious.testing.kafka.api;
 
+import java.time.Duration;
 import java.util.Map;
+import java.util.function.Predicate;
 
 /**
  * A KafkaCluster, from which is it possible to create/connect clients.
@@ -60,4 +62,9 @@ public interface KafkaCluster extends AutoCloseable {
      * @return mutable configuration map
      */
     Map<String, Object> getKafkaClientConfiguration(String user, String password);
+
+    /**
+     *
+     */
+    assertEachBrokerDescibesCluster(Predicate<org.apache.kafka.clients.admin.DescribeClusterResult> descibeClusterResultMatcher, Duration timeout);
 }
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
@@ -15,6 +15,7 @@ import kafka.server.KafkaServer;
 import kafka.server.Server;
 import kafka.tools.StorageTool;
 import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.admin.DescribeClusterResult;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.utils.Time;
 import org.apache.zookeeper.server.ServerCnxnFactory;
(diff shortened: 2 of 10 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -47,6 +52,7 @@ public class InVMKafkaCluster implements KafkaCluster {
     private final ZooKeeperServer zooServer;
     private final List<Server> servers;
     private final List<String> bootstraps = new ArrayList<>();
+    private final Supplier<KafkaClusterConfig.KafkaEndpoints> kafkaEndpointsSupplier;
```
SamBarker, line 55: I'm not entirely convinced this is the right place to keep this. 

It should maybe be stored in the clusterConfig and accessed via there.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -109,8 +114,7 @@ public EndpointPair getControllerEndpoint(int brokerId) {
 
             servers = clusterConfig.getBrokerConfigs(kafkaEndpointsSupplier, zookeeperEndpointSupplier).map(this::buildKafkaServer).collect(Collectors.toList());
 
-        }
-        catch (IOException e) {
+        } catch (IOException e) {
```
SamBarker, line 117: Must update my formatter settings to match the maven plug-in (thought I don't like the line break it adds)

api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
```diff
@@ -60,4 +62,9 @@ public interface KafkaCluster extends AutoCloseable {
      * @return mutable configuration map
      */
     Map<String, Object> getKafkaClientConfiguration(String user, String password);
+
+    /**
+     *
+     */
+    assertEachBrokerDescibesCluster(Predicate<org.apache.kafka.clients.admin.DescribeClusterResult> descibeClusterResultMatcher, Duration timeout);
```
k-wall, line 69: Is that actually used?

SamBarker: Not yet. 

I need to add a test to verify the behaviour.

k-wall:
Do we actually need this PR?  I have a memory that we agreed that it is the role of the test itself to await the cluster to come in coordination. IIRC the argument (expressed by @tombentley) was that the fact Kafka uses an eventually consistent model is inherent.  It would be wrong from a testing framework to hide that from test authors.

SamBarker:
> Do we actually need this PR? I have a memory that we agreed that it is the role of the test itself to await the cluster to come in coordination (https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/7#issuecomment-1337253584) . IIRC the argument (expressed by @tombentley) was that the fact Kafka uses an eventually consistent model is inherent

I think we have largely the same picture of what was agreed. I'm not trying to hide the distributed nature of Kafka however I do think the extension should ensure that the cluster is "ready" for use before handing it over to tests. I.e. if the cluster  is annotated with a broker count of `2` we should ensure `2` brokers are active prior to injection. Which is what I was trying to achieve here. Then provide some supporting code for more sensitive tests that actually want to ensure the coherence of the cluster.

k-wall:
> > Do we actually need this PR? I have a memory that we agreed that it is the role of the test itself to await the cluster to come in coordination ([#7 (comment)](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/7#issuecomment-1337253584)) . IIRC the argument (expressed by @tombentley) was that the fact Kafka uses an eventually consistent model is inherent
> 
> I think we have largely the same picture of what was agreed. I'm not trying to hide the distributed nature of Kafka however I do think the extension should ensure that the cluster is "ready" for use before handing it over to tests. I.e. if the cluster is annotated with a broker count of `2` we should ensure `2` brokers are active prior to injection. 

That was the view I held too. That's why I had initially add the flag file (being removed by #18)  stuff to the containers so they checked just that.  However, I think @tombentley arguments against that being the default behaviour are sound.  

> Which is what I was trying to achieve here. Then provide some supporting code for more sensitive tests that actually want to ensure the coherence of the cluster.

@tombentley point is that testing for cluster being complete should be **opt-in** at the test level.  Maybe a test annotation would be a unobtrusive way to achieve that? WDYT?

franvila:
> > > Do we actually need this PR? I have a memory that we agreed that it is the role of the test itself to await the cluster to come in coordination ([#7 (comment)](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/7#issuecomment-1337253584)) . IIRC the argument (expressed by @tombentley) was that the fact Kafka uses an eventually consistent model is inherent
> > 
> > 
> > I think we have largely the same picture of what was agreed. I'm not trying to hide the distributed nature of Kafka however I do think the extension should ensure that the cluster is "ready" for use before handing it over to tests. I.e. if the cluster is annotated with a broker count of `2` we should ensure `2` brokers are active prior to injection.
> 
> That was the view I held too. That's why I had initially add the flag file (being removed by #18) stuff to the containers so they checked just that. However, I think @tombentley arguments against that being the default behaviour are sound.
> 
> > Which is what I was trying to achieve here. Then provide some supporting code for more sensitive tests that actually want to ensure the coherence of the cluster.
> 
> @tombentley's point is that testing for cluster being complete should be **opt-in** at the test level. Maybe a test annotation would be a unobtrusive way to achieve that? WDYT? @franvila?

I agree with @tombentley. We should always avoid to modify the cluster behavior inside the test framework, all the steps needed should be included in each test, as not all kind of tests will need the same steps.

A good test principle to follow is always avoid assertions in the main code, the assertion should always be part of the test itself and it should be unique, that is there should be only one assertion in each test as it shall be the outcome of the test.

SamBarker:
I think by default if the user has requested a `2` broker cluster the extension should ensure there are `2` active brokers *before* it injects them, otherwise every test has to assert the framework has done its job before it can actually test what it is trying to do. Though I'm content to accept that that the framework doesn't need to ensure that the brokers are consistent and fully replicated.

There is possibly an argument for making allowing tests to opt out of that guarantee or a separate mechanism for creating clusters which are not fully functional.

SamBarker:
This PR was based around using `org.rnorth.duct-tape:duct-tape` in order to avoid a dependency on `org.awaitility:org.awaitility` which we now have anyway. I'm going to close this and revisit.

PR #15, opened 2022-12-22
Different kafka versions

### LINKED ISSUE #12: Add support for @Version
Some providers (e.g. container-based, or possibly forking, but not in-VM)  will be able to support multiple broker versions. We should try to implement this.

### LINKED ISSUE #19: CI should run tests in container mode too
Currently the CI exercises the test suite in VM mode only.   It should run in container mode too.

franvila:
- Added `@Version` parameter to select the version to be used in the test case with containers. `latest` by default (fix #12) 
- Added `@TestTemplate` for the new test cases
- Added container tests to the CI build as well (fix #19 )
- Added `KAFKA_IMAGE_REPO` and `ZOOKEEPER_IMAGE_REPO` environment variables to be able to select custom kafka and/or zookeeper images
- Selection of the proper quay image depending on the kafka version selected using `KAFKA_VERSION` environment variable (https://github.com/kroxylicious/kroxylicious/issues/86)


impl/src/test/java/io/kroxylicious/testing/kafka/common/KafkaClusterTest.java
```diff
@@ -67,30 +69,44 @@ public void kafkaClusterZookeeperMode() throws Exception {
         }
     }
 
-    @Test
-    public void kafkaTwoNodeClusterKraftMode() throws Exception {
-        int brokersNum = 2;
+    @TestTemplate
+    @ExtendWith(KafkaClusterTestInvocationContextProvider.class)
+    public void kafkaClusterTemplate(KafkaClusterTestCase testCase) throws Exception {
```
tombentley, line 74: The way the dependencies worked (in my head at least) was that the `KafkaClusterFactory`, `KafkaClusterConfig`  `KafkaClusterConfig` and related classes are lower level, and don't have a dependency on the annotations. This test was respecting that layering, but adding the extension mixes things up. Can we keep this test as it was, and add a new test, just for the `@Version` annotation?

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterTestInvocationContextProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+
+import static java.util.Arrays.asList;
+
+public class KafkaClusterTestInvocationContextProvider implements TestTemplateInvocationContextProvider {
```
tombentley, line 24: Why is this a separate thing from the existing extension? In particular, unless it is part of the existing extension I don't think it would be possible to write `@TestTemplate` methods which test the Cartesian product of a range of versions and `{kraft, zookeeper}`, or `{combined, dedicated}` KRaft nodes, or `{with SSL, without SSL}`.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Version.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.PARAMETER })
+public @interface Version {
+    String[] values();
```
tombentley, line 16: Should `@Version` support a range of values, or specify a single one? Compare with the other annotations where `@BrokerCluster` can't express a range of cluster sizes, for example. Instead we test over a range using `@DimensionMethodSource`, and by specifying multiple dimensions we test over the Cartesian product. 

The reason I did it like that is it's a bit more flexible. Specifying multiple values is fine when the single thing is a scalar, like version. But it doesn't work well you want to test over a range of cases where each case has to be specified as a more complex type such as sets of broker configs.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -144,6 +146,11 @@ public EndpointPair getControllerEndpoint(int brokerId) {
         }).collect(Collectors.toList());
     }
 
+    private void setDefaultKafkaImage(String kafkaVersion) {
+//        DEFAULT_KAFKA_IMAGE = DockerImageName.parse("quay.io/ogunalp/kafka-native:" + kafkaVersion + "-SNAPSHOT");
+        DEFAULT_KAFKA_IMAGE = DockerImageName.parse("quay.io/ogunalp/kafka-native:latest-snapshot");
```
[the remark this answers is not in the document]
franvila: This was commented only for testing purposes. Currently there is no kafkaVersion support on ogunalp repository. We are waiting for that to be done. That's why this PR is in draft mode.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Version.java
franvila: My intention was to set `@Version` as a parameter and then in the context provider use those range of values to launch the same tests for the versions indicated in the array. Inside the context provider I did different combinations with the brokersNum, KRaft/Zookeeper mode... as they were covered in the integration tests, but I can try to follow your approach also with the `@TestTemplate`

franvila: Done

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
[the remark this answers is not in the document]
franvila: done. Added a TODO comment into the version list of the KafkaClusterTest to be uncommented when this is ready on ozangunalp repository

impl/src/test/java/io/kroxylicious/testing/kafka/common/KafkaClusterTest.java
franvila: done

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterTestInvocationContextProvider.java
franvila: done

README.md
```diff
@@ -66,6 +67,15 @@ The following provisioning mechanisms are currently supported:
 Which kind of cluster is chosen depends on the requirements of your test.
 For example, using containers allows to easily test against different broker versions. 
 
+---
+**NOTE**
+
+In case you use podman for testcontainers, some tips must be taken into account:
+* `TESTCONTAINERS_RYUK_DISABLED=true` env variable shall be declared as per https://github.com/containers/podman/issues/7927#issuecomment-731525556
+* `podman-plugins` linux package shall be installed on your machine in order to communicate containers among each other.
```
SamBarker, line 75: Can we enforce this as part of the build? I'm happy for that to be in a separate PR.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -49,138 +63,134 @@ public class KafkaClusterTest {
...
-                .kraftMode(true)
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(1, cluster);
-        }
+    private static Stream<Version> versions() {
+        return Stream.of(
+                version("latest")
+        // TODO: waiting for new versions support in ozangunalp repo https://github.com/ozangunalp/kafka-native/issues/21
+        // version("3.3.1"),
```
SamBarker, line 72: Is there a version other than `latest` we could use now before https://github.com/ozangunalp/kafka-native/issues/21 is fixed?

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -49,138 +63,134 @@ public class KafkaClusterTest {
...
-                .brokersNum(brokersNum)
-                .kraftMode(true)
-                .build())) {
-            assumeTrue(cluster instanceof TestcontainersKafkaCluster, "KAFKA-14287: kraft timing out on shutdown in multinode case");
-            cluster.start();
-            verifyRecordRoundTrip(brokersNum, cluster);
-        }
+    private final Object[] brokerClusters = clusterSizes().toArray();
+
+    static Stream<List<Annotation>> tuples() {
```
SamBarker, line 85: can we give this a more descriptive name? `clusterDetails`? `clustersComposedOf`?

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -234,15 +244,13 @@ private void createTopic(AdminClient admin1, String topic, short replicationFact
...
-        this.testInfo = testInfo;
-        this.keytoolCertificateGenerator = new KeytoolCertificateGenerator();
-    }
-
-    @AfterEach
-    void after() {
-        Path filePath = Paths.get(keytoolCertificateGenerator.getCertLocation());
-        filePath.toFile().deleteOnExit();
+    @AfterAll
+    public static void cleanUp() throws IOException {
```
SamBarker, line 248: for completeness we should probably clear `clusterSizesIndex` here as well.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -49,138 +63,134 @@ public class KafkaClusterTest {
...
-        try (var cluster = KafkaClusterFactory.create(KafkaClusterConfig.builder()
-                .testInfo(testInfo)
-                .brokersNum(brokersNum)
-                .kraftMode(false)
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(brokersNum, cluster);
+    @BeforeEach
+    public void initializeIndexes() {
+        if (clusterSizesIndex == clusterSizes().count()) {
```
SamBarker, line 95: Why do we have an `if` guard here? Shouldn't it just reset it to zero for each test?

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -49,138 +63,134 @@ public class KafkaClusterTest {
...
+    }) TestcontainersKafkaCluster cluster)
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerZookeeperModeSSL(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @ZooKeeperCluster @Tls TestcontainersKafkaCluster cluster)
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
```
SamBarker, line 193: Not an issue directly from this PR but we should change https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/15/files#diff-98fb5d27cd329bdf14c1cf95c9163c9d859206844b4ba1d59d04b6a8d73672ddR207 so that it expects `message` rather than the hard-coded string.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: No, templates does not need to reset the indexes unless we have covered the total amount of `clusterSizes` dimension. It can be a little bit confusing but it is the only way I found to be able to get the values from `clusterSizes` to be checked inside the test.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: Not in the format we talked about. There is no image in quay pointing to a concrete kafka version as we expected, so we cannot currently point to any of the images following that approach

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: done

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: done

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
SamBarker: I'm struggling with the stateful nature that this gives for the tests. 

I'm starting to think the `KafkaCluster` should expose the broker count so that the tests can know how many brokers are involved.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
-            verifyRecordRoundTrip(brokersNum, cluster);
-        }
+    private final Object[] brokerClusters = clusterSizes().toArray();
+
+    static Stream<List<Annotation>> clusterSizesAndModes() {
+        return Stream.of(
+                List.of(brokerCluster(1), kraftCluster(1)),
+                List.of(brokerCluster(2), kraftCluster(1)),
+                List.of(brokerCluster(1), zooKeeperCluster()),
+                List.of(brokerCluster(2), zooKeeperCluster()));
```
SamBarker, line 86: I think I understand what this is doing, but it feels like a clunky way to express the problem. 

As I understand it we want to express a range of Kafka cluster sizes and a set of controller types and then generate the cartesian product of those two things. 

Which isn't quite what the parameter resolver does, at least I don't think that's what it does. I feel like we are working at the wrong level of abstraction because we manually duplicate the `kraftCluster` and the `zooKeeperCluster`. 

@tombentley what was your vision here? Am I missing something? 

Should we be using two `@DimensionMethodSource`s instead?

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
-                .securityProtocol("SASL_PLAINTEXT")
-                .saslMechanism("PLAIN")
-                .user("guest", "guest")
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(1, cluster);
-        }
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterTemplate(@ConstraintsMethodSource("clusterSizesAndModes") InVMKafkaCluster cluster)
```
SamBarker, line 98: Does the parameter type need to be the concrete `InVMKafkaCluster`? I would expect it to be the `KafkaCluster` interface. 

```suggestion
    public void kafkaClusterTemplate(@ConstraintsMethodSource("clusterSizesAndModes") KafkaCluster cluster)
```

SamBarker, line 98: As a test name this doesn't give me any clues what it's trying to demonstrate. So when I run the tests I get a nice sea of green but no idea what it has proved.

Screenshot 2023-01-17 at 11 20 05 AM

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
-                .securityProtocol("SASL_PLAINTEXT")
-                .saslMechanism("PLAIN")
-                .user("guest", "guest")
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(1, cluster);
-        }
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerKraftModeTemplate(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @KRaftCluster TestcontainersKafkaCluster cluster)
```
SamBarker, line 105:
```suggestion
    public void kafkaClusterContainerKraftModeTemplate(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @KRaftCluster KafkaCluster cluster)
```

Again not sure why we need the concrete type in the method signature.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
-                .securityProtocol("SASL_SSL")
-                .saslMechanism("PLAIN")
-                .user("guest", "guest")
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(1, cluster);
-        }
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerZookeeperModeTemplate(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @ZooKeeperCluster TestcontainersKafkaCluster cluster)
```
SamBarker, line 112:
```suggestion
    public void kafkaClusterContainerZookeeperModeTemplate(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @ZooKeeperCluster KafkaCluster cluster)
```

Again not sure why we need the concrete type in the method signature.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
-                .securityProtocol("SSL")
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(1, cluster);
-        }
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterWithAuth(@ConstraintsMethodSource("clusterSizesAndModes") @SaslPlainAuth({
+            @SaslPlainAuth.UserPassword(user = "guest", password = "guest")
+    }) InVMKafkaCluster cluster)
```
SamBarker, line 121:
```suggestion
    }) KafkaCluster cluster)
```

Again not sure why we need the concrete type in the method signature.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
-                .user("guest", "guest")
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(1, cluster);
-        }
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterSASL_SSL(@ConstraintsMethodSource("clusterSizesAndModes") @Tls @SaslPlainAuth({
+            @SaslPlainAuth.UserPassword(user = "guest", password = "guest")
+    }) InVMKafkaCluster cluster)
```
SamBarker, line 130:
```suggestion
    }) KafkaCluster cluster)
```

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
-                .keytoolCertificateGenerator(keytoolCertificateGenerator)
-                .kraftMode(false)
-                .securityProtocol("SSL")
-                .build())) {
-            cluster.start();
-            verifyRecordRoundTrip(1, cluster);
-        }
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterSSL(@ConstraintsMethodSource("clusterSizesAndModes") @Tls InVMKafkaCluster cluster)
```
SamBarker, line 137:
```suggestion
    public void kafkaClusterSSL(@ConstraintsMethodSource("clusterSizesAndModes") @Tls KafkaCluster cluster)
```

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerKraftModeWithAuth(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @KRaftCluster @SaslPlainAuth({
+            @SaslPlainAuth.UserPassword(user = "guest", password = "guest")
+    }) TestcontainersKafkaCluster cluster)
```
SamBarker, line 146:
```suggestion
    }) KafkaCluster cluster)
```

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerKraftModeSASL_SSL(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @KRaftCluster @Tls @SaslPlainAuth({
+            @SaslPlainAuth.UserPassword(user = "guest", password = "guest")
+    }) TestcontainersKafkaCluster cluster)
```
SamBarker, line 155:
```suggestion
    }) KafkaCluster cluster)
```

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerKraftModeSSL(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @KRaftCluster @Tls TestcontainersKafkaCluster cluster)
```
SamBarker, line 162:
```suggestion
    public void kafkaClusterContainerKraftModeSSL(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @KRaftCluster @Tls KafkaCluster cluster)
```

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerZookeeperModeWithAuth(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @ZooKeeperCluster @SaslPlainAuth({
+            @SaslPlainAuth.UserPassword(user = "guest", password = "guest")
+    }) TestcontainersKafkaCluster cluster)
```
SamBarker, line 171:
```suggestion
    }) KafkaCluster cluster)
```

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerZookeeperModeSASL_SSL(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @ZooKeeperCluster @Tls @SaslPlainAuth({
+            @SaslPlainAuth.UserPassword(user = "guest", password = "guest")
+    }) TestcontainersKafkaCluster cluster)
```
SamBarker, line 180:
```suggestion
    }) KafkaCluster cluster)
```

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -47,140 +59,134 @@ public class KafkaClusterTest {
...
+            throws Exception {
+        verifyRecordRoundTrip(((BrokerCluster) brokerClusters[clusterSizesIndex++]).numBrokers(), cluster);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void kafkaClusterContainerZookeeperModeSSL(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @ZooKeeperCluster @Tls TestcontainersKafkaCluster cluster)
```
SamBarker, line 187:
```suggestion
    public void kafkaClusterContainerZookeeperModeSSL(@DimensionMethodSource("versions") @DimensionMethodSource("clusterSizes") @ZooKeeperCluster @Tls KafkaCluster cluster)
```

SamBarker (commented):
I'm struggling with the fact that the test suite seems to be stateful between invocations of a `testTemplate`. 

Which isn't directly related to the `@Version` annotation but feels important to discuss and was surfaced by this PR.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: As `@Version` does not make sense for `IN_VM` tests, the only way I found to force to test in a container is using `TestcontainerKafkaCluster` instead of `KafkaCluster`. So, just in case, the other tests shall be `InVMKafkaCluster` to avoid any possible duplication

franvila: I can rename the test to show the goal of what we are testing, but the config if each test is pretty clear for me:
* We are testing in VM
* We know the number of brokers of each test
* We know if we are using zookeeper or KRaft mode

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: AFAIK if you run the test you can see the name of the tests telling the number of brokers used and all the info indicated in the `@DimensionMethodSource`

README.md
franvila: The env variable had been included by @k-wall here: https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/main/pom.xml#L260

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
SamBarker: I'm confused why doesn't `@Version` make sense in `IN_VM`? I can appreciate it being harder to orchestrate a specific version `IN_VM` but to me the requirement still makes sense. 

Also if we have tests which only make sense for one provisioning strategy I'd prefer them to be in their own file and explicitly configure the strategy, but that could just be me.

SamBarker: Yeah we can see what configuration the cluster is in but what are we behaviour are we asserting on the cluster? 

I view the tests as an executable specification and that most of their value comes from enhancing our understanding of the requirements. Just telling me what the configuration of a cluster is only part of the story.

franvila: AFAIK the kafka version for `IN_VM` comes from pom file, so it cannot be set dynamically inside the test, that's why I said that it does not make sense for that kind of tests.

We could use different files, it is not a problem from my side. WDYT @tombentley @k-wall ?

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
tombentley: @franvila it's not exactly obvious how to test that a `@TestTemplate` -annotated test is generating the set of tests you expect, but I did something a bit similar in `io.kroxylicious.testing.kafka.junit5ext.TemplateTest` which I think is a little cleaner:

* When a test is executed it adds to a map
* I do the assertion in an `@AfterAll`

It's not ideal, but I couldn't think of a nicer way without executing a test runner with an `@Test` which I felt was more mind bending.

Btw I still think that the `@TestTemplate` should be tested in `io.kroxylicious.testing.kafka.junit5ext.TemplateTest` rather than `KafkaClusterTest`. i.e. I still don't really understand why you've moved all these classes to `common`. In my mind the layering is:

```
common
  ^
  |
impl
  ^
  |
junit5
```

And that's because common and impl provide a useful API that doesn't depend on junit in any way. E.g. _in principal_ we could use the same classes to provide some similar functionality for an extension to some non-JUnit testing framework.

franvila: @tombentley I saw your tests in `io.kroxylicious.testing.kafka.junit5ext.TemplateTest` but the intention here is not testing the template itself, it is to make use of the template as a tool to perform an integration test and then in each test check if the kafka with the corresponding configuration is working fine or not. That's why I keep it in `KafkaClusterTest` class.

I have moved all classes to common because if I hadn't do that, we would have a circle dependency between `junit5` and `common`, that is I cannot see what is defined in `junit5` from `common` unless I add the dependency in the pom file what makes the circle dependency.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: I agree with you, but here I did not found a way to use `@DimensionMethodSource` that to switch between `kraftCluster` and `zookeeperCluster`, because returning an `Stream<Annotation>` does not work for `@DimensionMethodSource`.

So I decided to use the `@ConstraintsMethodSource` to avoid rewriting the test twice, one per each cluster mode.

My vision using the templates is a tool to reduce writing tests with `@Test` that test exactly the same but the configuration is different and can be done using the annotations.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: Reworked.

franvila: Reworked

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
franvila: Moved all template tests to `TemplateTest` class

franvila:
@tombentley @k-wall @SamBarker new structure has been uploaded since the conversation we have yesterday. Hope this makes more sense than the previous one. Any feedback is appreciated.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -126,6 +133,9 @@ public static KafkaClusterConfig fromConstraints(List<Annotation> annotations) {
             if (annotation instanceof ClusterId) {
                 builder.kafkaKraftClusterId(((ClusterId) annotation).value());
             }
+            if (annotation instanceof Version) {
+                builder.kafkaVersion(((Version) annotation).value());
```
SamBarker, line 137: Can their  & should there be some validation of versions here? 

Do we actually want to even suggest support for Kafka `0.11` for example? 

I'm quite happy to see version validation addressed as a separate PR.

SamBarker (approved):
I'm looking forward to seeing this actually work when the upstream issue is sorted.

LGTM

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
```diff
@@ -147,4 +150,27 @@ public static void checkTuples() {
...
+    @AfterAll
+    public static void checkVersions() {
+        assertEquals(Set.of("latest-snapshot"), observedVersions);
+    }
+
+    @TestTemplate
+    @ExtendWith(KafkaClusterExtension.class)
+    public void testVersions(@DimensionMethodSource("versions") @KRaftCluster TestcontainersKafkaCluster cluster)
+            throws Exception {
+        observedVersions.add(cluster.getKafkaVersion());
```
SamBarker, line 173: Why are we adding the observed version to a set and asserting the values in `afterAll` especially when there is only one version?

Even in a future when multiple versions are supported (I'd argue we should update this test when that becomes a reality rather than build for it now) the current implementation doesn't work. We would want to assert that every version generated by `versions` should be returned by `cluster.getKafkaVersion`.

franvila: Sorry but I don't see why it does not work for multiple versions.

I prefer to have it everything ready for the next change rather than leave it and having to make a big modification later on. But this is only my opinion.

SamBarker: It doesn't work as implemented anyway. The problem as I see it is as things stand you end up with duplicate definitions of versions one in the source method and one in the afterAll

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -144,6 +143,11 @@ public EndpointPair getControllerEndpoint(int brokerId) {
         }).collect(Collectors.toList());
     }
 
+    private void setDefaultKafkaImage(String kafkaVersion) {
+        DEFAULT_KAFKA_IMAGE = DockerImageName.parse("quay.io/ogunalp/kafka-native:" + kafkaVersion + "-snapshot");
```
k-wall, line 147: Let's extract "quay.io/ogunalp/kafka-native" as a constant.  We need some way to override that.  I *don't* think this needs to be exposed as an annotation. 

I think an environment variable or system property would be sufficient.

My use case are:
# I've just built kafka:next for the first time and it's in my own quay repo, I want to test it.
# I've hit this really knotty problem and I had to build a kafka image with extra instrumentation. I want an easy way to override it locally.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
k-wall: Making assertions in `afterAll` seems highly questionable. Let's not do that.

* using `afterAll` for assertions violates the Principle of Least Surprise.
* you are asserting the combined effect of the running of all the test, but how do you know the developer actually ran all the tests?  He could be passing invoking Test#runSingle or selecting tests from the IDE.  The failure will likely confuse him.

tombentley: I think @franvila is just re-using the pattern I established (above in the source code) for being able to assert that a `@TestTemplate`-annotated method really generated the expected the expected set of tests. I agree that making the assertion in `afterAll` is not normal. The reason I did it was because I couldn't find a better way of being able to make that assertion, using the facilities provided by junit jupiter, when testing a junit extension. 

Perhaps a better way, which might work, would be to move the test for `@TestTemplate` to the `integration-test` module, and do something like the `ExceptionalTest` for the templates.

franvila: That's right, I just tried to follow the same pattern established in the class. We could refactor this class if needed, but I think we should create another issue to fix that, because it is out of the scope of this PR, if you all agree.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
franvila: done

k-wall:
lgtm, thanks @franvila

PR #16, opened 2023-01-10
Capture container logs

k-wall:


SamBarker (approved):
Is it worth extracting a property so that `docker` users can run with `false`

k-wall:
> Is it worth extracting a property so that `docker` users can run with `false`

Done.

PR #23, opened 2023-01-19
Add new use case where client has no auth against SSL server

### LINKED ISSUE #20: Test framework should not assume SSL client auth
Currently the test harness enables SSL client auth whenever SSL is enable. It shouldn't be that way.  The use of SSL client should be optional and should be disabled by default.

https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/0d1066cc4f3d1d1558d72e34877205fad1a52984/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java#L231

franvila:
This change fixes #20 

@k-wall @tombentley Could you take a look at it?


impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -230,7 +230,7 @@ public Stream<ConfigHolder> getBrokerConfigs(Supplier<KafkaEndpoints> endPointCo
                 catch (GeneralSecurityException | IOException e) {
                     throw new RuntimeException(e);
                 }
-                server.put(BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG, "required");
+                server.put(BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG, "requested");
```
tombentley, line 233: Is this what you meant in #20 @k-wall? My interpretation wasn't that is was a complaint about the broker config, but that when we got a client for an SSL broker it was automatically and always already configured for SSL, rather than get client being configured for plain by default and getting an SSL client being opt-in.

franvila: AFAIK the client is always configured for PLAIN by default if you don't say in the test that you want to use TLS. That's why we need tests configured with `@Tls` (if we use the templates) annotation.

Is there any use case missing like having the broker configured with SSL and the client using no security at all? In that case, does it make sense? I read a lot of articles about kafka and no one recommends to use a plain client against an SSL broker (that's why I set it by default as `required`)

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -129,9 +131,11 @@ public void kafkaClusterZookeeperModeWithAuth() throws Exception {
 
     @Test
     public void kafkaClusterKraftModeSASL_SSL() throws Exception {
+        createClientCertificate();
         try (var cluster = KafkaClusterFactory.create(KafkaClusterConfig.builder()
                 .testInfo(testInfo)
-                .keytoolCertificateGenerator(keytoolCertificateGenerator)
+                .brokerKeytoolCertificateGenerator(brokerKeytoolCertificateGenerator)
+                .clientKeytoolCertificateGenerator(clientKeytoolCertificateGenerator)
```
k-wall, line 138: I would have a separate test(s) for the SSL client auth.  For the other SSL tests, the client should not have a keystore, only a trust store.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -220,23 +228,27 @@ public Stream<ConfigHolder> getBrokerConfigs(Supplier<KafkaEndpoints> endPointCo
...
                     throw new RuntimeException(e);
                 }
                 server.put(BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG, "required");
 
-                server.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, keytoolCertificateGenerator.getCertLocation());
-                server.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, keytoolCertificateGenerator.getPassword());
-                server.put(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, keytoolCertificateGenerator.getCertLocation());
-                server.put(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, keytoolCertificateGenerator.getPassword());
-                server.put(SslConfigs.SSL_KEY_PASSWORD_CONFIG, keytoolCertificateGenerator.getPassword());
+                server.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, brokerKeytoolCertificateGenerator.getTrustStoreLocation());
```
k-wall, line 247: THe broker only needs a trust store when SSL client auth is in use.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -284,12 +296,18 @@ public Map<String, Object> getConnectConfigForCluster(String bootstrapServers, S
...
+                catch (GeneralSecurityException | IOException e) {
+                    throw new RuntimeException(e);
+                }
+                kafkaConfig.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, clientKeytoolCertificateGenerator.getTrustStoreLocation());
+                kafkaConfig.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, clientKeytoolCertificateGenerator.getPassword());
                 if (securityProtocol.equals(SecurityProtocol.SSL.name())) {
-                    kafkaConfig.put(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, keytoolCertificateGenerator.getCertLocation());
-                    kafkaConfig.put(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, keytoolCertificateGenerator.getPassword());
-                    kafkaConfig.put(SslConfigs.SSL_KEY_PASSWORD_CONFIG, keytoolCertificateGenerator.getPassword());
+                    kafkaConfig.put(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, clientKeytoolCertificateGenerator.getKeyStoreLocation());
```
k-wall, line 308: The client only needs a keystore if SSL client auth is being used.

k-wall:
> Refactor done, @k-wall @tombentley Could you take a look at it?

The PR makes me think you haven't quite groked SSL client auth.

In the ordinary SSL case (that is without SSL client auth) it is **only** the server side that has a certificate/key pair.   
* It gives the client assurance that the server is who they say they are.
* The server has no knowledge who the client is.
* if the server cert is **self signed**, the client needs the server's cert **(not key)** in a truststore.  This is because the client's platform trust won't trust a self-signed cert.

In the SSL client auth case, **both** parties have their **own** certificate/key pair.
* it gives both client **and** server the assurance that the other side is who they say they are.
*  if the server cert is **self signed**, the client needs the server's cert **(not key)** in a truststore.
*  if the clients cert is **self signed**, the server needs the client's cert **(not key)** in a truststore.

You've made the code do the second case.   We want SSL client auth to be **optional**.

https://cheapsslsecurity.com/p/what-is-ssl-client-certificate-authentication-and-how-does-it-work/

franvila:
> > Refactor done, @k-wall @tombentley Could you take a look at it?
> 
> The PR makes me think you haven't quite groked SSL client auth.
> 
> In the ordinary SSL case (that is without SSL client auth) it is **only** the server side that has a certificate/key pair.
> 
> * It gives the client assurance that the server is who they say they are.
> * The server has no knowledge who the client is.
> * if the server cert is **self signed**, the client needs the server's cert **(not key)** in a truststore.  This is because the client's platform trust won't trust a self-signed cert.
> 
> In the SSL client auth case (aka mutually authenticated SSL or MASSL), **both** parties have their **own** certificate/key pair.
> 
> * it gives both client **and** server the assurance that the other side is who they say they are.
> * if the server cert is **self signed**, the client needs the server's cert **(not key)** in a truststore.
> * if the clients cert is **self signed**, the server needs the client's cert **(not key)** in a truststore.
> 
> You've made the code do the second case. We want SSL client auth to be **optional**.
> 
> https://cheapsslsecurity.com/p/what-is-ssl-client-certificate-authentication-and-how-does-it-work/

You are right! I forgot to add the first use case where the client has no certificate at all. I will do.

Thanks for your great feedback!

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -220,23 +228,27 @@ public Stream<ConfigHolder> getBrokerConfigs(Supplier<KafkaEndpoints> endPointCo
...
                 try {
-                    keytoolCertificateGenerator.generateSelfSignedCertificateEntry("test@redhat.com", clientEndpoint.getConnect().getHost(), "KI", "RedHat", null, null,
+                    brokerKeytoolCertificateGenerator.generateSelfSignedCertificateEntry("test@redhat.com", clientEndpoint.getConnect().getHost(), "KI", "RedHat", null, null,
                             "US");
+                    brokerKeytoolCertificateGenerator.generateTrustStore(clientKeytoolCertificateGenerator.getCaCertFilePath(), clientEndpoint.getConnect().getHost());
                 }
                 catch (GeneralSecurityException | IOException e) {
                     throw new RuntimeException(e);
                 }
                 server.put(BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG, "required");
```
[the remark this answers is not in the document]
franvila: Now the SSL_CLIENT_AUTH_CONFIG is set to `requested`, that means that the client can use or not TLS to connect to the server.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
franvila: done

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
franvila: done

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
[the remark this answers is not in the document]
k-wall: "requested" isn't the same.  If it is "requested" the server will use the client's certificate it it is provided.  That's actually a different use-case.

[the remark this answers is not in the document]
franvila: refactored

k-wall:
thanks @franvila .  The KeytoolCertificateGenerator API doesn't seem quite right to me, but this was an existing issue.  I vote we merge this and move one.

@SamBarker WDYT?

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
```diff
@@ -27,24 +26,42 @@
...
+        this.trustStoreFilePath = (trustStorePath != null) ? Path.of(trustStorePath) : Paths.get(certsDirectory.toAbsolutePath().toString(), "kafka.truststore.jks");
+
+        certsDirectory.toFile().deleteOnExit();
+        if(certFilePath == null) {
+            this.keyStoreFilePath.toFile().deleteOnExit();
         }
-        catch (IOException e) {
-            throw new UncheckedIOException(e);
+        if(trustStorePath == null) {
+            this.trustStoreFilePath.toFile().deleteOnExit();
```
SamBarker, line 48: nit: I'm confused why we test if the `trustStorePath` is null and then mark `this.trustStoreFilePath` as delete on exit.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
```diff
@@ -77,17 +120,31 @@ public void generateSelfSignedCertificateEntry(String email, String domain, Stri
...
+    private void createCrtFileToImport() throws IOException {
+        final List<String> commandParameters = new ArrayList<>(List.of("openssl", "pkcs12"));
+        commandParameters.addAll(List.of("-in", getKeyStoreLocation()));
+        commandParameters.addAll(List.of("-passin", "pass:" + getPassword()));
+        commandParameters.add("-nokeys");
+        commandParameters.addAll(List.of("-out", certFilePath.toAbsolutePath().toString()));
+        runCommand(commandParameters);
+    }
+
+    private void runCommand(List<String> commandParameters) throws IOException {
```
SamBarker, line 147: Oh yikes we can't use key tool via a native API? We have to invoke it as an external process.

SamBarker (approved):
Surprised to see us invoking keytool as a process and not as a java API but that is certainly out of scope for this PR to address.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
franvila: Because in the constructor the kafka client can set its own trustStorePath and we won't delete this file. If `null` we are sure that the trustStoreFilePath is temporary and we can delete it.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
franvila: I have investigated this and the native API is not public any more since Java 9 I believe, I would need to look for that info again to be sure from which version it is not available.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
SamBarker: It's probably worth a comment to that effect, given the odds on me remembering that are slim...

PR #33, opened 2023-02-24
Fix #28: probe each broker within the cluster individually when assessing readiness

### LINKED ISSUE #28: TemplateTest#checkCartesianProduct test flakey
I'm seeing this maybe 50% of the runs.

```
TESTCONTAINERS_RYUK_DISABLED=true mvn verify -Dmaven.javadoc.skip=true
```

io.kroxylicious.testing.kafka.junit5ext.TemplateTest  Time elapsed: 15.068 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <[[3, snappy], [1, snappy], [3, zstd], [1, zstd]]> but was: <[[1, zstd], [1, snappy], [2, snappy], [3, zstd]]>

k-wall:
Stabilizing  tests that use more than one broker.

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
@@ -17,14 +17,19 @@ import java.io.UncheckedIOException;
 import java.net.InetAddress;
 import java.net.ServerSocket;
 import java.time.Duration;
-import java.util.Collection;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.function.Predicate.not;
+import static org.apache.kafka.clients.CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG;
 import static org.slf4j.LoggerFactory.getLogger;
 
 public class Utils {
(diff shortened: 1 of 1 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -53,28 +58,54 @@ public static Stream<ServerSocket> preAllocateListeningSockets(int num) {
...
+        while(knownReady.size() < expectedBrokerCount) {
+            if (toProbe.isEmpty()) {
+                throw new IllegalArgumentException(String.format("Ran out of addresses to probe before cluster ready: %s", originalBootstrap));
+            }
+            var probeAddress = toProbe.iterator().next();
+
+            var copy = new HashMap<>(connectionConfig);
+            copy.put(BOOTSTRAP_SERVERS_CONFIG, probeAddress);
+
+            try (Admin admin = Admin.create(copy)) {
```
SamBarker, line 76: Feels odd to be re-creating the admin client every iteration.

I haven't checked what happens if it fails to connect in one iteration what happens on a subsequent one.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -53,28 +58,54 @@ public static Stream<ServerSocket> preAllocateListeningSockets(int num) {
...
-                        catch (TimeoutException te) {
-                            log.warn("Kafka timed out describing the the cluster");
-                        }
-                        return Collections.emptyList();
-                    }, Matchers.hasSize(expectedBrokerCount));
+        var knownReady = Collections.synchronizedSet(new HashSet<String>());
+        var toProbe = Collections.synchronizedSet(new HashSet<String>());
+
+        var originalBootstrap = String.valueOf(connectionConfig.get(BOOTSTRAP_SERVERS_CONFIG));
+        toProbe.addAll(Arrays.asList(originalBootstrap.split(",")));
```
SamBarker, line 65: If we are going to probe each broker independently I would prefer to make that explicit in the api and pass each address separately rather than reversing them back out of the bootstrap string.

Having said that in the spirit of progress over polish, I'd be happy to do that in a separate PR.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -53,28 +58,54 @@ public static Stream<ServerSocket> preAllocateListeningSockets(int num) {
...
+            }
+            var probeAddress = toProbe.iterator().next();
+
+            var copy = new HashMap<>(connectionConfig);
+            copy.put(BOOTSTRAP_SERVERS_CONFIG, probeAddress);
+
+            try (Admin admin = Admin.create(copy)) {
+                Awaitility.await()
+                        .pollDelay(Duration.ZERO)
+                        .pollInterval(1, TimeUnit.SECONDS)
```
SamBarker, line 79: Seeing this again makes me think we should configure the awaitility defaults rather than customising the specific call, but that's not really part of this PR's scope.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -53,28 +58,54 @@ public static Stream<ServerSocket> preAllocateListeningSockets(int num) {
...
+        var knownReady = Collections.synchronizedSet(new HashSet<String>());
+        var toProbe = Collections.synchronizedSet(new HashSet<String>());
+
+        var originalBootstrap = String.valueOf(connectionConfig.get(BOOTSTRAP_SERVERS_CONFIG));
+        toProbe.addAll(Arrays.asList(originalBootstrap.split(",")));
+
+        while(knownReady.size() < expectedBrokerCount) {
+            if (toProbe.isEmpty()) {
+                throw new IllegalArgumentException(String.format("Ran out of addresses to probe before cluster ready: %s", originalBootstrap));
+            }
```
SamBarker, line 70: Thinking out loud would this read better as a compound while condition? `knownReady.size() < expectedBrokerCount && toProbe.iterator().hasNext()` with a check after the loop to make sure everything was in sync and present?

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -53,28 +58,54 @@ public static Stream<ServerSocket> preAllocateListeningSockets(int num) {
...
+                        .pollDelay(Duration.ZERO)
+                        .pollInterval(1, TimeUnit.SECONDS)
+                        .atMost(timeout, timeUnit)
+                        .ignoreExceptions()
+                        .until(() -> {
+                            log.info("describing cluster: {}", probeAddress);
+                            try {
+                                admin.describeCluster().controller().get().id();
+                                var nodes = admin.describeCluster().nodes().get(10, TimeUnit.SECONDS);
+                                log.info("got nodes: {}", nodes);
```
SamBarker, line 87: Probably should be debug. We should also log the probe address so we can tell which nodes have which members.

SamBarker (commented):
I want to try running this locally before approving (reviewing from my iPad) but change looks sane to me.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
k-wall: > Feels odd to be re-creating the admin client every iteration.
I suspect that we could move the admin client outside the retry loop.  I don't feel an imperative to change tbh

SamBarker: Each `create` causes a [new thread](https://github.com/apache/kafka/blob/5c52c61a46f103cda7fa48bd96584dafe984d6d3/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L596) to be created which is less than ideal.

SamBarker (approved):
I think the change is good, but it's not improving the reliability of the test suite for me. 

The TemplateTest is still failing with intermittent reasons related to consistency

k-wall:
> I think the change is good, but it's not improving the reliability of the test suite for me.
> 
> The TemplateTest is still failing with intermittent reasons related to consistency

Like this https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/36?

SamBarker:
Yes, like 36 which I had put down to metadata inconsistencies, but I think I had other random failures too

PR #40, opened 2023-03-01
Update formatting

SamBarker:
Configures GitHub to fail builds where the formatting is wrong to prevent merging without the agreed formatting.


robobario (approved):
LGTM, tested the build is failed locally with `-Pci` and formatting errors

PR #44, opened 2023-03-02
Disable JMX metrics

k-wall:
The test harness doesn't rely upon Kafka JMX metrics (and probably won't because JMX isn't supported by the kafka native). Registering/Unregistering the mbeans is time-consuming so we disable it.

No functional change.


SamBarker (approved):
LGTM

Issue #36, opened 2023-02-28
TemplateTest#testTuples flakey


k-wall:
```
[ERROR] Tests run: 13, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 17.38 s <<< FAILURE! - in io.kroxylicious.testing.kafka.junit5ext.TemplateTest
[ERROR] io.kroxylicious.testing.kafka.junit5ext.TemplateTest  Time elapsed: 17.38 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <[[1, 1], [3, -1], [3, 1]]> but was: <[[1, 1], [3, -1]]>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
```

k-wall:
I wonder if it has to do with the order listeners are coming up.   Currently the code uses TreeSets, keyed by listener name so ANON is first in the listener list.  The Kafka code kicks off the listeners in order, but *I think* the code is async, so which listener comes ready first is non-deterministic.

k-wall:
No, it isn't listener order causing the problem.  Even if I probe EXTERNAL, I still see. the issue.


2023-03-01 14:05:58 WARN  io.kroxylicious.testing.kafka.invm.InVMKafkaCluster:158 - Generated config {node.id=2, offsets.topic.num.partitions=1, offsets.topic.replication.factor=1, listener.security.protocol.map=ANON:PLAINTEXT,CONTROLLER:PLAINTEXT,EXTERNAL:PLAINTEXT,INTERNAL:PLAINTEXT, listeners=ANON://0.0.0.0:61626,EXTERNAL://0.0.0.0:61623,INTERNAL://0.0.0.0:61629, group.initial.rebalance.delay.ms=0, inter.broker.listener.name=INTERNAL, log.dir=/var/folders/5s/v5576rp175bf4vrf15mgvffm0000gn/T/kafka7019848818931926552/broker-2, controller.quorum.voters=0@//localhost:61630, controller.listener.names=CONTROLLER, process.roles=broker, broker.id=2, advertised.listeners=ANON://localhost:61626,EXTERNAL://localhost:61623,INTERNAL://localhost:61629, early.start.listeners=ANON,EXTERNAL,INTERNAL}
[...]

PR #45, opened 2023-03-02
Don't list external or anon as early.start.listeners

k-wall:
"A comma-separated list of listener names which may be started before the authorizer has finished initialization.... A listener should not appear in this list if it accepts external traffic."

I noticed this whilst investigating #36. It is not related, not it seems worthwhile fixing.


SamBarker (approved):
LGTM

PR #46, opened 2023-03-02
Allow TemplateTest methods to be run individually

k-wall:
Fix a couple of logging levels too.

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactory.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactory.java
@@ -11,7 +11,7 @@ import io.kroxylicious.testing.kafka.api.KafkaCluster;
 import io.kroxylicious.testing.kafka.invm.InVMKafkaCluster;
 import io.kroxylicious.testing.kafka.testcontainers.TestcontainersKafkaCluster;
 
-import static java.lang.System.Logger.Level.INFO;
+import static java.lang.System.Logger.Level.DEBUG;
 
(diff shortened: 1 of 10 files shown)


SamBarker (commented):
This is largely my ignorance but why does making the shared state fields `AtomicReferences` achieve the goal from the PR title.

franvila:
Me too, it is the first time I see `AtomicReference` as well and don't know how it works at all.

franvila (commented):
Me too, it is the first time I see `AtomicReference` as well and don't know how it works at all.

k-wall:
Problem I am trying to solve: It is bothersome that we cannot run a test individually -Dtest=TemplateTest#testTuples without seeing a whole bunch of spurious test failures corresponding to tests that haven't run.

Maybe the ARs were distateful/magic.  I was relying on the compare and swap setting the AR once if the test actually ran and making the afterall execution check that it had (with an assumeTrue).

So I had another go at achieving the same ends.   Now I'm using Nested tests which means you can runs the tests in the IDE, or from the command line.

```
mvn verify -Dmaven.javadoc.skip=true '-Dtest=TemplateTest$Tuples#testTuples'
```

Alternatively, using a separate file per test case is an option, but that would mean moving the static methods into their own file `clusterSizes` etc is shared.    I don't like that as that would separate the test data from its assertion,.

SamBarker (approved):
LGTM I had played with a similar change but didn't twig the idea of having the nested class implement `AfterAllCallback`

franvila:
LGTM. Thanks @k-wall !

PR #70, opened 2023-03-30
Configure `__transactionState_` topic like we do `__consumer_offsets_`

SamBarker:
To support running tests requiring transactions we need to configure the `__transactionState_` topic. Hard coded to `1` for now even if multiple brokers are supported as we don't need the replication.


robobario (approved):
LGTM

k-wall (approved):

PR #77, opened 2023-04-06
Release script

SamBarker:
Add a script for running releases of the extension.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,84 @@
+#
+# Copyright Kroxylicious Authors.
+#
+# Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+#
+
+set -e
+
+REPOSITORY="origin"
+BRANCH_FROM="main"
+while getopts ":v:b:r:k:" opt; do
+  case $opt in
+    v) RELEASE_VERSION="${OPTARG}"
+    ;;
+    b) BRANCH_FROM="${OPTARG}"
+    ;;
+    r) REPOSITORY="${OPTARG}"
+    ;;
+    k) GPG_KEY="${OPTARG}"
+    ;;
+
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -0,0 +1,95 @@
+#
```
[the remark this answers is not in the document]
k-wall: Sam fixed this already.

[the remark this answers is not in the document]
SamBarker: Oh I was hoping the bot would auto resolve it.

scripts/release.sh
```diff
@@ -0,0 +1,96 @@
...
+
+if [[ -z ${RELEASE_VERSION} ]]; then
+  echo "No version specified aborting"
+  exit 1
+fi
+
+GREEN='\033[0;32m'
+NC='\033[0m' # No Color
+
+git stash --all
```
k-wall, line 51: I've normally favoured re-cloning the repo to a temporary location and running the release from there in order to eliminate the risk of pollution from a developer's own workspace.  You are doing a git stash, but what if I have local commits? What do you think?

k-wall (approved):
I've not executed the script myself, but it all looks good to me.  thank you @SamBarker

scripts/release.sh
SamBarker: I didn't think too deeply but I thought `stash` plus branch from `remote/branch` would avoid the local commits issue. See [#L55](https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/77/files#diff-1bba00e5aca52286a667c09eff92ba2ca8e3ca77e0d31b0599cc829ab0173389R55) 

Do you have a sample of how you've managed a clean checkout in the past?

k-wall: I don't have any example directly to hand.  I didn't notice the remote/branch.  You approach is good enough.

PR #82, opened 2023-04-14
Added getNumOfBrokers method in KafkaCluster interface

### LINKED ISSUE #56: Expose cluster size thru KafkaCluster
Suppose I'm writing a test method which uses a `KafkaCluster`. I might want to write an assumption on the cluster size. But I can't so this easily today because `KafkaCluster` doesn't expose the expected size of the cluster. 

You could argue that you should know the size of the `KafkaCluster` you asked for (via `@BrokerCluster`), but sometimes it's more convenient to get this information at the use site, rather than the declaration size.

franvila:
- Added the getNumOfBrokers method into the KafkaCluster interface that close #56
- Refactor some tests that make use of the new method
- Added new readme section to validate the format of the files

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,3 +223,18 @@ This will produce tests as follows:
 Provisioning mechanisms can be provided externally by implementing `io.kroxylicious.testing.kafka.api.KafkaClusterProvisioningStrategy` and declaring it as a [Java service](https://www.baeldung.com/java-spi) (e.g. in a  `META-INF/services/io.kroxylicious.testing.kafka.api.KafkaClusterProvisioningStrategy` file that is available on the test-time classpath).
 
 You can also provide your own constraint annotation types by annotating them with the `@io.kroxylicious.testing.kafka.api.KafkaClusterConstraint` meta-annotation. Such custom constraint annotations will only be understood by your custom provisioning strategy, so the in-JVM and testcontainers-based clusters provided by this project then can't be used.
+
+## Compilation with format validator
+
+As the project should follow the same format for all the files, in case a format error is detected when running the `ci` profile:
+
+```shell
+mvn clean verify -Pci
+```
+
+the following commands shall be run to fix those format issues: 
+
+```shell
+mvn compile
+mvn formatter:validate
+```
--- a/api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
+++ b/api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
@@ -29,7 +29,7 @@ public interface KafkaCluster extends AutoCloseable {
     void close() throws Exception;
 
     /**
-     *Returns the size of the cluster.
+     * @return the size of the cluster.
      */
     int getNumOfBrokers();
 
--- a/api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
+++ b/api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
@@ -29,6 +29,7 @@ public interface KafkaCluster extends AutoCloseable {
     void close() throws Exception;
 
     /**
+     * Gets the number of brokers expected in the cluster
      * @return the size of the cluster.
      */
     int getNumOfBrokers();
@@ -40,6 +41,7 @@ public interface KafkaCluster extends AutoCloseable {
     String getBootstrapServers();
 
     /**
+     * Gets the cluster id
      * @return The cluster id for KRaft-based clusters, otherwise null;
      */
     String getClusterId();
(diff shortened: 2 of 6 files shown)


api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
```diff
@@ -29,7 +29,7 @@ public interface KafkaCluster extends AutoCloseable {
     void close() throws Exception;
 
     /**
-     * @return the size of the cluster
+     *Returns the size of the cluster.
```
SamBarker, line 32: I have to say I think lift is just wrong here. 

We should keep the `@return` doc as well.

README.md
```diff
@@ -223,3 +223,18 @@ This will produce tests as follows:
 Provisioning mechanisms can be provided externally by implementing `io.kroxylicious.testing.kafka.api.KafkaClusterProvisioningStrategy` and declaring it as a [Java service](https://www.baeldung.com/java-spi) (e.g. in a  `META-INF/services/io.kroxylicious.testing.kafka.api.KafkaClusterProvisioningStrategy` file that is available on the test-time classpath).
 
 You can also provide your own constraint annotation types by annotating them with the `@io.kroxylicious.testing.kafka.api.KafkaClusterConstraint` meta-annotation. Such custom constraint annotations will only be understood by your custom provisioning strategy, so the in-JVM and testcontainers-based clusters provided by this project then can't be used.
+
+## Compilation with format validator
+
+As the project should follow the same format for all the files, in case a format error is detected when running the `ci` profile:
```
SamBarker, line 229: The maven build defaults to fixing formatting errors. The `ci` profile is purely intended so that the build fails if the formatting is wrong so that a human validates the fixes.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
```diff
@@ -48,46 +46,20 @@ static Stream<BrokerCluster> clusterSizes() {
...
+                                         @DimensionMethodSource(value = "clusterSizes", clazz = TemplateTest.class) KafkaCluster cluster)
+            throws ExecutionException, InterruptedException {
+        try (var admin = Admin.create(cluster.getKafkaClientConfiguration())) {
+            assertEquals(admin.describeCluster().nodes().get().size(), cluster.getNumOfBrokers());
         }
     }
 
-    static Map<Integer, Integer> observedMultipleClusterSizesWithAdminParameters = new HashMap<>();
-
-    @Nested
```
SamBarker, line 75: Why are we loosing the `@Nested` annotation here?

SamBarker (approved):
LGTM

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
franvila: It is not needed anymore as we can assert in each iteration of the template and not waiting till the end of the loop to assert in the `@AfterAll` section that honestly I don't like very much

README.md
franvila: Agree with that. But it happens to me that when testing locally without `ci` profile everything seems to be fine, but on github action that uses `ci` profile, the build fails because of that. I had to do it manually following the steps I put in the README file.

This is a suggestion just in case someone has the same issue than me.

api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
franvila: Agree, changed by sonatype

api/src/main/java/io/kroxylicious/testing/kafka/api/KafkaCluster.java
```diff
@@ -28,6 +28,11 @@ public interface KafkaCluster extends AutoCloseable {
     @Override
     void close() throws Exception;
 
+    /**
+     * @return the size of the cluster.
```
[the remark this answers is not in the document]
franvila: @sonatype-lift ignore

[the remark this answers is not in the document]
franvila: @sonatype-lift ignore

README.md
SamBarker: Hmmm, that suggests something else is going wrong for you locally. One shouldn't need to run those commands `mvn install` should fix it up anyway.

I'm happy to keep the hints there but would like to work out why we don't just do the right thing by default.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
SamBarker: Ahh Sweet, That's great!

PR #87, opened 2023-04-20
Remove hardcoded key ID from mvn gpg plugin config

grdryn:
The change here is basically copied from the main kroxylicious repo.


robobario (approved):
LGTM

PR #88, opened 2023-04-21
Kroxylicious junit extension Release v0.2.0 2023-04-21

racheljpg:
Release version 0.2.0

PR #89, opened 2023-04-21
Kroxylicious junit extension development version 2023-04-21

racheljpg:
prepare for new development version

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -14,7 +14,7 @@
     <parent>
         <groupId>io.kroxylicious.testing</groupId>
         <artifactId>testing-parent</artifactId>
-        <version>0.2.0-SNAPSHOT</version>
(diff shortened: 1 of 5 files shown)


grdryn:
```

grdryn:
I'm not sure why the release script creates 2 divergent commits and two separate PRs for each (this and #88). It's possible to accidentally merge them in the wrong order, which would lead to a wrong state. Actually I think merging either of them will lead to a conflict in the other, since they're editing exactly the same lines in different directions (from a git graph perspective).

I'm not sure what's better: should the script put the commits ordered on one PR branch and just create one PR, or should we just not bother having the release commit on the main branch (with regards to the main branch, it's an intermediate state that just exists during the release process (and it stored at the tag), right?)? I think the former. I'll try to propose a change to the release script to show what I mean.

PR #91, opened 2023-04-24
Only make one PR after release

grdryn:
Previously 2 commits would be made on different branches from main, and a PR created against main on GitHub. Merging either of those would cause a conflict in the other. Also, there's a specific order that those commits need to merged in, so having two pull requests increases the potential for human error.

The intention of this change is to make it so that there will only be one PR, with the two commits ordered as they should be.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -76,21 +76,17 @@ git push --tags
 echo "Deploying release to maven central"
 mvn deploy -Prelease -DskipTests=true -DreleaseSigningKey="${GPG_KEY}"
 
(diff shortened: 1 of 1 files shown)


grdryn:
This is related to https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/89#issuecomment-1518046998

k-wall (approved):
I've not run the changes, but your idea for a single PR makes sense to me.

SamBarker (approved):
I had it as two separate PR's as I viewed them as two independent states and didn't want to conflate them, but I agree that it wasn't a great flow or experience. 

The other aspect I wanted to inject in the middle between the two version commits was:
1. Creating tags
2. Creation of the release page in GH.

Both of which can happen post the combined merge so I'm happy to go with it.

PR #94, opened 2023-04-27
Retry INVM server startup.

SamBarker:
Startup is cheap, and is usually related to transient port binding issues. 

I was hoping to say we could drop this once we get to Kafka 3.6.0 but then I saw https://github.com/apache/kafka/pull/13572#issuecomment-1516694321 so there is some debate if it will land and when...

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
@@ -13,6 +13,7 @@ import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -182,7 +184,11 @@ public void start() {
             }
         }
 
-        servers.stream().parallel().forEach(Server::startup);
+        servers.stream().parallel().forEach(server -> Awaitility.await().atMost(Duration.ofSeconds(5)).pollDelay(Duration.ofMillis(50)).until(() -> {
```
robobario, line 187: Wondering if 5 seconds is a bit tight for slow machines

SamBarker: I was hesitant to go longer as its all happening in cluster setup so could be quite slow for extension users. However a working cluster is probably more important than speed so will bump it up.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -21,30 +22,33 @@
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.utils.Time;
 import org.apache.zookeeper.server.ServerCnxnFactory;
 import org.apache.zookeeper.server.ZooKeeperServer;
 import org.jetbrains.annotations.NotNull;
 
-import io.kroxylicious.testing.kafka.api.KafkaCluster;
```
k-wall, line 29: formatter-maven-plugin is reverting this change.

k-wall (approved):
Change looks good.  I reviewed the `kafka.server.KafkaRaftServer#startup` and `kafka.server.KafkaServer#startup` to ensure they have idempotency.  In the failure, they both take steps to call shutdown down methods which appear to reverse any parts of startup that succeeded.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
SamBarker: Argh! Will fix. 

Need to have another go at syncing the config

franvila:
Great workaround for this kind of failure! LGTM. Thanks @SamBarker

franvila (approved):

PR #103, opened 2023-05-10
Fixed the password set in the configuration cluster config

### LINKED ISSUE #97: @SaslPlainAuth.UserPassword password not respected
When using a `@Test` method parameter declared like this:

```java
@SaslPlainAuth.UserPassword(user="foo", password = "pa55word"))
KafkaCluster kc
```

The code in `KafkaClusterConfig` incorrectly uses the user name as the password 

https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/bfaad781e0a036211cde9656eafd458907c2e52a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java#L395

(the 3rd argument should be a call to `getValue()` rather than `getKey()`)

franvila:
Fixes #97 

Updated as well the test cases which use the same user and password and avoided the issue to be found.


tombentley (approved):
Thanks!

PR #104, opened 2023-05-11
Added target to saslPlainAuth userpassword interface

### LINKED ISSUE #98: Missing `@Target` on `SaslPlainAuth.UserPassword`
The annotation type `SaslPlainAuth.UserPassword` is missing a `@java.lang.annotation.Target` annotation on its declaration. 
This allows users to write `@Test` methods which type check but where the annotated `KafkaCluster` parameter is not actually configured for SASL as they might expect. For example:

```java
@Test
void myTest(@SaslPlainAuth.UserPassword(user = "oo", password = "") KafkaCluster cluster)  {
```

This can be fixed by annotating the declaration of `@interface UserPassword` with `@Target({})`. It will still be allowed as the `value` of an enclosing `@SaslPlainAuth` annotation instantiation.

franvila:
- Added the target to the SaslPlainAuth UserPassword interface
- Added new test case to cover that new case

Fixes #98

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -103,6 +103,7 @@ public class KafkaClusterConfig {
             KRaftCluster.class,
             Tls.class,
             SaslPlainAuth.class,
+            SaslPlainAuth.UserPassword.class,
             ZooKeeperCluster.class,
             Version.class);
 
@@ -154,6 +155,11 @@ public class KafkaClusterConfig {
                         .collect(Collectors.toMap(
                                 SaslPlainAuth.UserPassword::user,
                                 SaslPlainAuth.UserPassword::password)));
+            }
+            if (annotation instanceof SaslPlainAuth.UserPassword) {
+                builder.saslMechanism("PLAIN");
+                sasl = true;
+                builder.users(Map.of(((SaslPlainAuth.UserPassword) annotation).user(), ((SaslPlainAuth.UserPassword) annotation).password()));
             }
             if (annotation instanceof ClusterId) {
                 builder.kafkaKraftClusterId(((ClusterId) annotation).value());
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/SaslPlainAuth.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/SaslPlainAuth.java
@@ -32,6 +32,9 @@ public @interface SaslPlainAuth {
     /**
      * The interface User password.
      */
+    @Target({ ElementType.FIELD, ElementType.PARAMETER })
+    @Retention(RetentionPolicy.RUNTIME)
+    @KafkaClusterConstraint
     @interface UserPassword {
         /**
          * Gets the username
--- a/junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
+++ b/junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
@@ -142,7 +142,7 @@ public class ParameterExtensionTest extends AbstractExtensionTest {
     }
 
     @Test
-    public void saslPlainAuthenticatingClusterParameter(
+    public void saslPlainAuthenticatingClusterParameter2Users(
                                                         @BrokerCluster @SaslPlainAuth({
                                                                 @SaslPlainAuth.UserPassword(user = "alice", password = "foo"),
                                                                 @SaslPlainAuth.UserPassword(user = "bob", password = "bar")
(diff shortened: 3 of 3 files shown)


tombentley (commented):
This solves the issue in a different way than stated in the issue. In the issue I was assuming we'd stick to the existing model where you annotate the cluster with:

```java
@SaslPlainAuth({@UserPassword(), @UserPassword})
```

i.e. `UserPassword` should not allowed as a top level annotation (achieved by annotating it with `@Target({})`), and should only be used as the value of a top level `@SaslPlainAuth` annotation. 

What this PR does is allow `@UserPassword` as a top level annotation. 

There is a difference in terms of what the API can express. For example at some point we might want to support a SCRAM-SHA-256 authentication. Do we want to support multiple authentication mechanisms on the same `KafkaCluster`, and if so should we support different credentials for different mechanisms? The answers inform the annotations. 

* If we only support one auth mechanism maybe we should have annotations like 
`@Scram(type=PLAIN, users={@User(name="tom", password="foo")})`
* If we only support multiple mechanisms maybe we should have annotations like 
`@Scram(type={PLAIN, SCRAM_256}, users={@User(name="tom", password="foo")})`

etc.

franvila:
This is the only way I have found to make this work
```
@SaslPlainAuth.UserPassword(user="XXX", password="XXX") KafkaCluster cluster
```
Setting only the `@Target{}` annotation in UserPassword as stated in the issue does not work because of the provision strategy used in the framework so I allowed to use `SaslPlainAuth.UserPassword()` as a supported constraint.

The issue states: 
> This can be fixed by annotating the declaration of @interface UserPassword with @Target({})

That's why I tried to follow this approach.

impl/src/main/java/io/kroxylicious/testing/kafka/common/SaslPlainAuth.java
```diff
@@ -32,6 +32,9 @@
     /**
      * The interface User password.
      */
+    @Target({ ElementType.FIELD, ElementType.PARAMETER })
+    @Retention(RetentionPolicy.RUNTIME)
+    @KafkaClusterConstraint
```
k-wall, line 37: I haven't experience of meta annotations, but `@ @KafkaClusterConstraint` looks wrong.  I think we want to express a constraint that says this annotation may only appear within the @SaslPlainAuth annotation.  So I think you want to introduce another kind of constraint check.

franvila: I agree, I'm a bit lost here, but If I don't set this constraint I cannot use this kind of annotation as requested:
```
@SaslPlainAuth.UserPassword(user="XXX", password="XXX") KafkaCluster cluster
```
In the `KafkaClusterConfig` class it never creates the map for the `users` because it is not recognize as a proper annotation for the `KafkaCluster`.

k-wall:
Not for this PR, but of the two ideas expressed above, I think there is a third:

> If we only support one auth mechanism maybe we should have annotations like
> @Scram(type=PLAIN, users={@User(name="tom", password="foo")})
> If we only support multiple mechanisms maybe we should have annotations like
> @Scram(type={PLAIN, SCRAM_256}, users={@User(name="tom", password="foo")})

How about `@Sasl` being `@Repeatable`?

impl/src/main/java/io/kroxylicious/testing/kafka/common/SaslPlainAuth.java
tombentley: did you try?

```java
@SaslPlainAuth{users={@UserPassword(user="XXX", password="XXX")}) KafkaCluster cluster
```

tombentley:
Technically the SCRAM-SHA annotations could support additional methods today (`iterations` for example). I'm not saying we _need_ that. And while I don't expect any new mechanisms it's not exactly certain that they'd conform to username and password and nothing else. If you used a single annotation type and some mechanisms need extra or different properties then you end up with the union of their properties in the annotation type and it's all a bit of a mess.

franvila:
@k-wall @tombentley I have made it `Repeatable`, modified the tests to make use of the new config and it works fine.
WDYT?

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -147,13 +147,19 @@ public static KafkaClusterConfig fromConstraints(List<Annotation> annotations) {
                     throw new RuntimeException(e);
                 }
             }
-            if (annotation instanceof SaslPlainAuth) {
+            if (annotation instanceof SaslPlainAuth.List) {
```
k-wall, line 150: you can use JEP 305 (instanceof pattern matching)

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -147,13 +147,19 @@ public static KafkaClusterConfig fromConstraints(List<Annotation> annotations) {
...
-                        .collect(Collectors.toMap(
-                                SaslPlainAuth.UserPassword::user,
-                                SaslPlainAuth.UserPassword::password)));
+                Map<String, String> users = new HashMap<>();
+                for (var user : ((SaslPlainAuth.List) annotation).value()) {
+                    users.put(user.user(), user.password());
+                }
+                builder.users(users);
+            }
+            else if (annotation instanceof SaslPlainAuth) {
```
k-wall, line 159: ditto JEP 305.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -103,6 +102,7 @@ public class KafkaClusterConfig {
             KRaftCluster.class,
             Tls.class,
             SaslPlainAuth.class,
```
k-wall, line 104: Do we need both SaslPlainAuth.class and SaslPlainAuth.List.class?  Isn't a single SaslPlainAuth just a list of one?

franvila: I followed the approach that it was. We had an array of `UserPassword` and a test case with more than one user/password so that's why I did this List. It is the same to the `BrokerConfig.class` and `BrokerConfig.List.class` as well

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
franvila: Done

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
franvila: done

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: From experimentation, I see you are right.

k-wall (approved):

k-wall:
> > This change breaks the API, so we should really bump the major when it is released.
> 
> Totally agree. It is breaking change for old tests

As we are still in the [0.x phase](https://semver.org/#spec-item-4), my comments about bumping the major are spurious.

PR #108, opened 2023-05-15
Fix #84: allow brokers to be added to/removed from the cluster

k-wall:
Extends the `KafkaCluster` API to introduce the following functionality:

* `addBroker`
* `removeBroker`

The methods are blocking so won't return until the broker is added or removed.  I decided my original idea of returning Futures didn't really fit in the existing API or fit well with likely use-cases (the test would probably just block until the Future completed).

Issue #86, opened 2023-04-18
API to restart kafka brokers


k-wall:
Thinking forward about testing Kroxylicious edge cases - network disconnection.

As the author of a component like Kroxylicious, I want to be able to write integration tests that ensure my component correctly handles broker restarts correctly.  To this end I would value an API that permitted me to restart one or more brokers within a cluster.  It is important that the broker comes back on the same ports (as it would in production).  I have this requirement for both the in-VM and container use-case.

I'm imagine a API that looks like:

`#restart(int nodeId)`
or maybe 
`#restart(Set<Integer> nodeId)`

control would return to the caller once the  broker restart is finished and all the brokers see one and other.

I don't envisaged needing functionality to do things like roll the brokers of a cluster (waiting for ISRs etc).  If a test needs that, it could easily construct it itself.

I don't envisage needing the ability to restart any kafka broker that is solely in the kraft controller role.

k-wall:
@SamBarker @racheljpg @devguyio @franvila @robobario  comments/edits welcome..

tombentley:
> all the brokers see one and other.

This is actually tricky to do using the Admin client. You have no control over which brokers get sent the `Metadata`/`DescribeCluster` request, nor do you know for sure that the metadata you receive isn't from before the restart. If might be necessary to observe the broker logging to do this robustly, if you're wedded to this semantic.

k-wall:
@tombentley so the approach that @SamBarker and I landed on was to [create a single partition topic with `RF = <number of brokers>`](https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/864531d765127d425d2d02f9e61961214cf0a5ca/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java#L121) and wait for that to be reported as having replicas on all brokers.   That seems to have stabilized the tests.

k-wall:
@franvila wondering if you'd have capacity to take a look at this?   being able to demonstrate kroxy's robustness will be key to first production use-cases.

k-wall:
I'm picking this feature up.

PR #109, opened 2023-05-17
Refactor KafkaEndpoints / KafkaCluster interaction

k-wall:
This PR is based on #108, please don't review those commits (the first three) on this PR.


impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -131,74 +146,38 @@ public TestcontainersKafkaCluster(DockerImageName kafkaImage, DockerImageName zo
...
-                kafkaContainer.dependsOn(this.zookeeper);
-            }
-            return kafkaContainer;
-        }).collect(Collectors.toList());
+    @NotNull
+    private KafkaContainer buildKafkaContainer(KafkaClusterConfig.ConfigHolder holder) {
+        String netAlias = "broker-" + holder.getBrokerNum();
+        KafkaContainer kafkaContainer = new KafkaContainer(kafkaImage)
+                .withName(name)
+                .withNetwork(network)
```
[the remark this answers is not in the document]
k-wall: @sonatype-lift exclude file

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -246,21 +232,22 @@ public String getKafkaVersion() {
...
-    private Stream<GenericContainer<?>> allContainers() {
+    private synchronized Stream<GenericContainer<?>> allContainers() {
         return Stream.concat(
-                this.brokers.stream(),
+                this.brokers.values().stream(),
                 Stream.ofNullable(this.zookeeper));
     }
 
     @Override
     @SneakyThrows
```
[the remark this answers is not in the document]
k-wall: @sonatype-lift exclude issue

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -116,7 +184,7 @@ private static KafkaFuture<Void> createTopic(Integer expectedBrokerCount, Admin
                     log.debug("Create topic future completed.");
                     if (throwable != null) {
                         log.warn("Failed to create topic: {} due to {}", CONSISTENCY_TEST, throwable.getMessage(), throwable);
-                        if (throwable instanceof RetriableException) {
+                        if (throwable instanceof RetriableException || throwable instanceof InvalidReplicationFactorException) {
                             CompletableFuture.supplyAsync(() -> createTopic(expectedBrokerCount, admin));
```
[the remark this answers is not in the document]
k-wall: @sonatype-lift ignore

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -302,6 +352,40 @@ public Map<String, Object> getKafkaClientConfiguration(String user, String passw
         return clusterConfig.getConnectConfigForCluster(getBootstrapServers(), user, password);
```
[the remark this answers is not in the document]
k-wall: @sonatype-lift ignore

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -245,12 +250,76 @@ public Map<String, Object> getKafkaClientConfiguration(String user, String passw
         return clusterConfig.getConnectConfigForCluster(getBootstrapServers(), user, password);
```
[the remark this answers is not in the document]
k-wall: @sonatype-lift ignore

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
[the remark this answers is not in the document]
k-wall: @sonatype-lift ignore

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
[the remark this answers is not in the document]
k-wall: @sonatype-lift ignore

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -330,7 +331,22 @@ public synchronized void removeBroker(int nodeId) throws UnsupportedOperationExc
         ports.values().forEach(pm -> pm.remove(nodeId));
 
         var kafkaContainer = brokers.remove(nodeId);
-        kafkaContainer.stop();
+        // Note that GenericContainer#stop actually implements stop using kill, so the broker doesn't have chance to
```
k-wall, line 334: Need to raise a TC issue about this.

robobario:
One thing I noticed was the state/allocation logic was looking quite similar over InVM and TestContainers, with quite a few boilerplate usages of the Preallocator. I've had a go at extracting a common component to tidy it up. Here's a [PR](https://github.com/k-wall/kroxylicious-junit5-extension/pull/1) raised against your fork. WDYT?

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -85,9 +106,14 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster {
         }
     }
 
-    private final KafkaClusterConfig.KafkaEndpoints kafkaEndpoints;
-    private List<ServerSocket> clientPorts;
-    private List<ServerSocket> anonPorts;
+    /**
+     * Protected by lock of {@link TestcontainersKafkaCluster itself.}
+     */
+    private final SortedMap<Integer, ServerSocket> clientPorts = new TreeMap<>();
```
robobario, line 112: `clientPorts` and `anonPorts` are dead code

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -75,8 +81,23 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster {
...
+     */
+    private final Map<Integer, KafkaContainer> brokers = new TreeMap<>();
+
+    /**
+     * Tracks ports that are in-use by this cluster, by listener.
+     * The inner map is a mapping of kafka <code>node.id</code> to a closed server socket, with a port number
+     * previously defined from the ephemeral range.
+     * Protected by lock of {@link TestcontainersKafkaCluster itself.}
+     */
+    private final Map<Listener, SortedMap<Integer, ServerSocket>> ports = Map.of(Listener.EXTERNAL, new TreeMap<>(),
```
robobario, line 99: With the various changes to the port pre-allocation, do we need to pass around `ServerSocket` or should port allocation return int?

robobario (approved):
LGTM with nits (including [this PR](https://github.com/k-wall/kroxylicious-junit5-extension/pull/1))

robobario (approved):
Whoops, meant to approve #108, had already approved this one

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -245,12 +244,75 @@ public Map<String, Object> getKafkaClientConfiguration(String user, String passw
...
+        try (var preallocator = new ListeningSocketPreallocator()) {
+            ports.get(Listener.EXTERNAL).put(newNodeId, preallocator.preAllocateListeningSockets(1).get(0));
+            ports.get(Listener.ANON).put(newNodeId, preallocator.preAllocateListeningSockets(1).get(0));
+            ports.get(Listener.INTERNAL).put(newNodeId, preallocator.preAllocateListeningSockets(1).get(0));
+        }
+
+        var configHolder = clusterConfig.generateConfigForSpecificBroker(this, newNodeId);
+        final Server server = buildKafkaServer(configHolder);
+        tryToStartServerWithRetry(configHolder, server);
+        servers.put(configHolder.getBrokerNum(), server);
```
SamBarker, line 269: I think we should really use `newNodeId` here or change `configHolder.getBrokerNum` to `configHolder.getBrokerId`

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -131,79 +147,55 @@ private KafkaConfig buildBrokerConfig(KafkaClusterConfig.ConfigHolder c, Path te
...
-        })
-                .collect(Collectors.toList());
-
-        Utils.awaitExpectedBrokerCountInClusterViaTopic(clusterConfig.getAnonConnectConfigForCluster(kafkaEndpoints), 120, TimeUnit.SECONDS,
+            tryToStartServerWithRetry(configHolder, server);
+            servers.put(configHolder.getBrokerNum(), server);
+
+        });
+        Utils.awaitExpectedBrokerCountInClusterViaTopic(
+                clusterConfig.getConnectConfigForCluster(buildServerList(brokerId -> getEndpointPair(Listener.ANON, brokerId)), null, null, null, null), 120,
```
SamBarker, line 167: Why did you get rid of the wrapper methods which hid this forest of `null` args?

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -245,12 +244,75 @@ public Map<String, Object> getKafkaClientConfiguration(String user, String passw
...
     @SuppressWarnings("ResultOfMethodCallIgnored")
     @Override
-    public void close() throws Exception {
+    public synchronized void close() throws Exception {
         try {
             try {
-                servers.stream().parallel().forEach(Server::shutdown);
+                // shutdown controller last seems to be significant to the kraft case.
+                servers.entrySet().stream()
+                        .sorted(Map.Entry.comparingByKey(Comparator.reverseOrder()))
```
SamBarker, line 313: I'm guessing this works because the controller is just the first broker started (assuming we use combined mode for the broker & controller)? Should we be doing something to more explicitly track which node(s) are controllers and make sure they are shut down after?

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -272,25 +333,17 @@ public void close() throws Exception {
...
-            }
-        });
+    @Override
+    public synchronized EndpointPair getEndpointPair(Listener listener, int nodeId) {
+        var portMap = ports.get(listener);
+        if (portMap == null) {
+            throw new IllegalStateException("Unexpected value: " + listener);
+        }
+        var port = portMap.get(nodeId);
+        return EndpointPair.builder().bind(new Endpoint("0.0.0.0", port.getLocalPort())).connect(new Endpoint("localhost", port.getLocalPort())).build();
```
SamBarker, line 347: This makes me wonder if it should be the EndPoints we keep in the nested map of the ports map.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -264,8 +374,8 @@ private void produce(KafkaCluster cluster, String topic, String message) throws
                 ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class,
                 ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class,
                 ProducerConfig.DELIVERY_TIMEOUT_MS_CONFIG, 3_600_000));
-        try (var producer = new KafkaProducer<String, String>(config)) {
-            producer.send(new ProducerRecord<>(topic, "my-key", message)).get();
+        try (var producer = CloseableProducer.wrap(new KafkaProducer<String, String>(config))) {
```
SamBarker, line 377: Does make me wonder if we should add a ctor to `Closeable*` which accepts `config` and creates the `Kafka*` internally.

SamBarker (approved):
Thanks @k-wall a few things to tidy up but looks good.

franvila (approved):
LGTM, all the findings in #108 has been fixed. Thanks @k-wall!

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -40,6 +50,64 @@ public class Utils {
...
+
+        try (var admin = Admin.create(connectionConfig)) {
+            awaitCondition(timeout, timeUnit).until(() -> {
+
+                Map<String, TopicDescription> topicDescriptions = describeKnownTopics(kafkaInternalTopics, admin);
+                var movements = new HashMap<TopicPartition, Optional<NewPartitionReassignment>>();
+                var toNodeReassignment = Optional.of(new NewPartitionReassignment(List.of(toNodeId)));
+                topicDescriptions.forEach((name, description) -> {
+                    // find all partitions that don't have a replica on at least one other node.
+                    var toMove = description.partitions().stream().filter(p -> p.replicas().stream().anyMatch(n -> n.id() == fromNodeId) && p.replicas().size() < 2)
```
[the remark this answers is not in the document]
k-wall: My rationale here was to only assign internal topic partitions that don't have a replica on another node.  Kafka itself should take care of transferring the leadership of those where replicas > 1 (this should take place during the [graceful shutdown  ](https://kafka.apache.org/documentation/#basic_ops_restarting)).   

As `[generateConfigForSpecificBroker](https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/beaa0f59f535beb6c703235c9c97bf6d28d3ccc3/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java#L311)` configures the internal topics to have one replica, we'd normally expect to follow our "reassign" path.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -131,79 +147,55 @@ private KafkaConfig buildBrokerConfig(KafkaClusterConfig.ConfigHolder c, Path te
...
-                            return false;
-                        }
-                    });
-            return server;
-        })
-                .collect(Collectors.toList());
-
-        Utils.awaitExpectedBrokerCountInClusterViaTopic(clusterConfig.getAnonConnectConfigForCluster(kafkaEndpoints), 120, TimeUnit.SECONDS,
+            tryToStartServerWithRetry(configHolder, server);
+            servers.put(configHolder.getBrokerNum(), server);
```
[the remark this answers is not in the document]
k-wall: Let's hold on this idea. #86 wants introduce the ability to restart broker(s). For that you'd expect the broker to come back on the same ports.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: I think `node.id` is the new parlance.  I think Kafka adopted it because with KRaft nodes aren't necessarily in the `broker` role, so `broker.id` would have been confusing.  I think it makes sense for us to adopt the new term.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: that was unintentional, I'll fix that.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: > I'm guessing this works because the controller is just the first broker started (assuming we use combined mode for the broker & controller)?

that's it.

It feels like we are hitting a kraft bug.  It seems `parallel()` was actually masking it.  `parallel()` is a no-op for streams derived from a treemap so that is how it was uncovered.  I'll see if I can put together a kafka only reproducer (separate from this work).

> Should we be doing something to more explicitly track which node(s) are controllers and make sure they are shut down after?

We have the information we need to do this already. I'll make a change.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: Yeah, agreed,  I'll have a poke at that and see how it looks.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
k-wall: Yeah, I'll do that.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
k-wall: It already exists: https://github.com/testcontainers/testcontainers-java/issues/1000

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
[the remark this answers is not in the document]
SamBarker: I wasn't intending to suggest reallocating ports at this stage and would expect to only do it in the face of a bind exception but I agree there is no major reason to change yet.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
k-wall: I added create factory methods as that seem most in keeping with the existing API.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -347,9 +308,14 @@ public synchronized void removeBroker(int nodeId) throws IllegalArgumentExceptio
     public synchronized void close() throws Exception {
         try {
             try {
-                // shutdown controller last seems to be significant to the kraft case.
+                // with kraft, if we don't shut down the controller last, we sometimes see a hang.
+                var controllers = ports.get(Listener.CONTROLLER).keySet();
```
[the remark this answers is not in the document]
k-wall: @sonatype-lift ignore

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -131,79 +147,55 @@ private KafkaConfig buildBrokerConfig(KafkaClusterConfig.ConfigHolder c, Path te
...
+    public synchronized void start() {
         // kraft mode: per-broker: 1 external port + 1 inter-broker port + 1 controller port + 1 anon port
         // zk mode: per-cluster: 1 zk port; per-broker: 1 external port + 1 inter-broker port + 1 anon port
         try (var preallocator = new ListeningSocketPreallocator()) {
-            externalPorts = preallocator.preAllocateListeningSockets(clusterConfig.getBrokersNum());
-            anonPorts = preallocator.preAllocateListeningSockets(clusterConfig.getBrokersNum());
-            interBrokerPorts = preallocator.preAllocateListeningSockets(clusterConfig.getBrokersNum());
-            controllerPorts = allocateControllerPorts(clusterConfig, preallocator);
+            List.of(Listener.EXTERNAL, Listener.ANON, Listener.INTERNAL)
+                    .forEach(l -> Utils.putAllListEntriesIntoMapKeyedByIndex(preallocator.preAllocateListeningSockets(clusterConfig.getBrokersNum()), ports.get(l)));
```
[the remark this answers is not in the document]
k-wall: I've merged Rob's PR into this PR, which has killed the repetitious mappery

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: With Rob's changes, I'm reached the 'that's good enough' point on this PR.  I'm inclined not to action this now, unless others feel strongly.

franvila (approved):
LGTM, but could you add javadoc headers to the public methods to avoid future warnings in the compilation?

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -347,9 +284,14 @@ public synchronized void removeBroker(int nodeId) throws IllegalArgumentExceptio
...
-                // shutdown controller last seems to be significant to the kraft case.
+                // with kraft, if we don't shut down the controller last, we sometimes see a hang.
+                // https://issues.apache.org/jira/browse/KAFKA-14287
                 servers.entrySet().stream()
-                        .sorted(Map.Entry.comparingByKey(Comparator.reverseOrder()))
+                        .filter(e -> !portsAllocator.containsPort(Listener.CONTROLLER, e.getKey()))
+                        .map(Map.Entry::getValue)
+                        .forEach(Server::shutdown);
+                servers.entrySet().stream()
+                        .filter(e -> portsAllocator.containsPort(Listener.CONTROLLER, e.getKey()))
```
SamBarker, line 294: I think this filter condition is extraneous as calling shutdown on the non controllers is safe (AFICT).

SamBarker (approved):
LGTM thanks @k-wall 

Some minor comments left.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
SamBarker: I wonder if we should actually do what they suggest in https://github.com/testcontainers/testcontainers-java/issues/1000#issuecomment-840657453 and call graceful stop ourselves rather than killing? (Entirely happy for that to be a separate PR)

impl/src/main/java/io/kroxylicious/testing/kafka/common/PortAllocator.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public synchronized int getPort(Listener listener, int nodeId) {
+        var listenerPorts = ports.get(listener);
+        if (listenerPorts == null) {
+            throw new IllegalArgumentException("listener " + listener + " has not been registered with this allocator.");
+        }
+        var sock = listenerPorts.get(nodeId);
+        if (sock == null) {
+            throw new IllegalArgumentException("listener " + listener + " does not have a port for node " + nodeId + ".");
+        }
```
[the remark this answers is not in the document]
k-wall: Given our usage pattern is to allocate a set of ports for a range of node-ids, if we try to get one that doesn't exist, something has gone wrong. Preferring fail early.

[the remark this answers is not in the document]
k-wall: refactored to call containsPort

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: I wasn't so sure about the idempotency in the kraft case.   If I invoke `kafka.server.KafkaRaftServer#shutdown` twice, what stops the second invocation overtaking the first?   In our case it would be safe as we are calling our shutdowns sequentially from a single thread.  To be safe I'd prefer to keep it this way.

impl/src/main/java/io/kroxylicious/testing/kafka/common/PortAllocator.java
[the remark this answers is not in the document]
SamBarker: Yeah I guess, allocating here possibly won't end up wiring into all the appropriate places.

We can revisit if we see the exception cropping up

PR #117, opened 2023-05-29
added await condition to wait for the consistency test topics to be deleted

franvila:
- Added a loop for all active brokers to wait for the topic to be deleted only when `delete.topic.enable` is not `false` (there is a test that is set to `false` on purpose and it causes troubles because it never deletes the topic).

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
@@ -20,12 +20,14 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.clients.admin.Config;
 import org.apache.kafka.clients.admin.NewPartitionReassignment;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.TopicDescription;
 import org.apache.kafka.common.KafkaFuture;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.TopicPartitionInfo;
+import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.errors.InvalidReplicationFactorException;
 import org.apache.kafka.common.errors.RetriableException;
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
@@ -151,6 +153,14 @@ public class Utils {
                         final Boolean isQuorate = promise.getNow(false);
                         if (isQuorate) {
                             admin.deleteTopics(Set.of(CONSISTENCY_TEST));
+                            for (int i = 0; i < admin.describeCluster().nodes().get().size(); i++) {
+                                ConfigResource resource = new ConfigResource(ConfigResource.Type.BROKER, String.valueOf(i));
+                                Config configs = admin.describeConfigs(List.of(resource)).all().get().get(resource);
+                                if (!configs.get("delete.topic.enable").value().equals("false")) {
+                                    awaitCondition(timeout, timeUnit)
+                                            .until(() -> !admin.listTopics().names().get().contains(CONSISTENCY_TEST));
+                                }
+                            }
                         }
                         return isQuorate;
                     });


impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -151,6 +153,14 @@ public static void awaitExpectedBrokerCountInClusterViaTopic(Map<String, Object>
                         final Boolean isQuorate = promise.getNow(false);
                         if (isQuorate) {
                             admin.deleteTopics(Set.of(CONSISTENCY_TEST));
+                            for (int i = 0; i < admin.describeCluster().nodes().get().size(); i++) {
```
SamBarker, line 156: I think we should have timeouts on the `get` call. I'm also kinda wary of doing a remote RPC on every loop.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -151,6 +153,14 @@ public static void awaitExpectedBrokerCountInClusterViaTopic(Map<String, Object>
...
+                                ConfigResource resource = new ConfigResource(ConfigResource.Type.BROKER, String.valueOf(i));
+                                Config configs = admin.describeConfigs(List.of(resource)).all().get().get(resource);
+                                if (!configs.get("delete.topic.enable").value().equals("false")) {
```
SamBarker, line 159: I think we should move the if test around the whole delete shouldn't we? There is no point doing any of this if `delete.topic.enable == false`.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -151,6 +153,14 @@ public static void awaitExpectedBrokerCountInClusterViaTopic(Map<String, Object>
...
+                                    awaitCondition(timeout, timeUnit)
+                                            .until(() -> !admin.listTopics().names().get().contains(CONSISTENCY_TEST));
+                                }
+                            }
```
SamBarker, line 163: Why don't we just block until the futures in [`DeleteTopicsResult.topicNameValues`](https://kafka.apache.org/32/javadoc/org/apache/kafka/clients/admin/DeleteTopicsResult.html#topicNameValues()) complete? Thought I think that should probably a separate awaitCondition rather than being nested inside the one that creates the topic.

SamBarker (changes requested):
@franvila I'm not sure we need the loop as I think we can just wait for the delete topic future to complete. 

If I'm wrong about that I think there is some cleanup to be done with the current implementation.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
franvila: If we remove this if clause, the test `ParameterExtensionTest.brokerConfigs` fails, because it forces to set that condition `delete.topic.enable == false` and it will never delete the topic so we will receive an exception when the timeout ends.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
franvila: I'm not an expert on futures, but can we wait for the completion only if the condition `delete.topic.enable` is not false?

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
SamBarker: I meant move the test for the config property to be before the attempt to delete so that we don't try to delete if its set to false.

franvila: Yeap, sorry I misunderstood the point of your comment. Totally agree

SamBarker: All good 😀

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
franvila: I don't know why if I wait for the future to complete I got always these errors:
```
2023-05-30 07:30:22 ERROR state.change.logger:74 - [Broker id=2] Received LeaderAndIsrRequest with correlation id 1 from controller 0 epoch 2 for partition __org_kroxylicious_testing_consistencyTest-0 (last update controller epoch 1) but cannot become follower since the new leader -1 is unavailable.
2023-05-30 07:30:22 ERROR state.change.logger:74 - [Broker id=0] Received LeaderAndIsrRequest with correlation id 1 from controller 0 epoch 2 for partition __org_kroxylicious_testing_consistencyTest-0 (last update controller epoch 1) but cannot become follower since the new leader -1 is unavailable.
2023-05-30 07:30:22 ERROR state.change.logger:76 - [Controller id=0 epoch=2] Controller 0 epoch 2 initiated state change of replica 2 for partition __org_kroxylicious_testing_consistencyTest-0 from OnlineReplica to ReplicaDeletionIneligible failed
```
But if I do the `awaitCondition` it never happens

SamBarker: Either awaitCondition is swallowing the errors or running in a background thread allows the deletion to happen properly.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
franvila: I have removed the loop. It does not make sense if we remove the topics in the cluster and wait for the removal for each broker

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
franvila: I tried to leave as it was, deleting the topics anyway. In that way we don't modify the behavior of the deletion action (if kafka is implemented well, the deletion won't be done if `delete.topic.enable == false`) and then we wait for the deletion only when needed.

SamBarker: I'm sure Kafka will do the right thing, however I was looking to avoid the RPC call if we knew it wouldn't work. Having said that given we need to make another RPC call to work that out its possibly moot.

I'll have a play tomorrow my time with blocking on the future even if its just to understand why one gets the errors you observed.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -155,7 +155,7 @@ public static void awaitExpectedBrokerCountInClusterViaTopic(Map<String, Object>
                             admin.deleteTopics(Set.of(CONSISTENCY_TEST));
                             ConfigResource resource = new ConfigResource(ConfigResource.Type.BROKER, "0");
                             Config configs = admin.describeConfigs(List.of(resource)).all().get().get(resource);
-                            if (!configs.get("delete.topic.enable").value().equals("false")) {
+                            if (configs != null && !configs.get("delete.topic.enable").value().equals("false")) {
```
SamBarker, line 158: grr Kafka really needs a `getOrDefault`...

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
franvila: Right. I put it inside the if clause.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
SamBarker: I've had a play with it and reproduced your error. 

1.  Create topic
2. Topic replicated
3. Call delete topic
3. 1. `delete.topic.enable==true` delete completes normally. 
3. 2. `delete.topic.enable==false` delete fails with exception. `awaitCondition(timeout, timeUnit)` will retry the consistency check. The topic is deleted/marked for deletion and thus can't be correctly described and causes an exception and fails the await.

I've opened  [#2](https://github.com/franvila/kroxylicious-junit5-extension/pull/2) to wait for the future and handle the `delete.topic.enable==false`. 

I've opted to let the delete fail as I think that optimises for the common case where `delete.topic.enable==true`.

franvila: Thanks @SamBarker!

franvila: I still got the error:
```
2023-05-31 07:33:00 ERROR state.change.logger:74 - [Broker id=0] Received LeaderAndIsrRequest with correlation id 1 from controller 0 epoch 2 for partition __org_kroxylicious_testing_consistencyTest-0 (last update controller epoch 1) but cannot become follower since the new leader -1 is unavailable.
```

SamBarker: I'll need to double check my terminal history... 

My change got the tests passing again

franvila: The tests always pass but the error appears and it is not fine to have an error that appears from time to time. It should be controlled to avoid future test errors

SamBarker: I think a passing suite that logs an error is better than an intermittent suite.

However I would rather not trigger the error

SamBarker: Running locally I get:

```
2023-06-01 10:56:21 awaitility-thread-309 INFO  kafka.server.KafkaRaftServer:66 - [KafkaRaftServer nodeId=0] Kafka Server started
2023-06-01 10:56:21 main-1 DEBUG io.kroxylicious.testing.kafka.common.Utils:123 - Creating topic: __org_kroxylicious_testing_consistencyTest via localhost:25859
2023-06-01 10:56:21 main-1 DEBUG io.kroxylicious.testing.kafka.common.Utils:127 - Waiting for __org_kroxylicious_testing_consistencyTest to be replicated to 1 brokers
2023-06-01 10:56:21 awaitility-thread-386 DEBUG io.kroxylicious.testing.kafka.common.Utils:130 - Calling describe topic
2023-06-01 10:56:21 EventHandler-360 INFO  kafka.server.ReplicaFetcherManager:66 - [ReplicaFetcherManager on broker 0] Removed fetcher for partitions Set(__org_kroxylicious_testing_consistencyTest-0)
2023-06-01 10:56:21 kafka-admin-client-thread | adminclient-12-385 DEBUG io.kroxylicious.testing.kafka.common.Utils:206 - Create topic future completed.
2023-06-01 10:56:21 kafka-admin-client-thread | adminclient-12-385 DEBUG io.kroxylicious.testing.kafka.common.Utils:145 - Current topicDescriptions: {__org_kroxylicious_testing_consistencyTest=(name=__org_kroxylicious_testing_consistencyTest, internal=false, partitions=(partition=0, leader=localhost:25859 (id: 0 rack: null), replicas=localhost:25859 (id: 0 rack: null), isr=localhost:25859 (id: 0 rack: null)), authorizedOperations=null)}
2023-06-01 10:56:21 kafka-admin-client-thread | adminclient-12-385 DEBUG io.kroxylicious.testing.kafka.common.Utils:193 - Expected number of replicas found.
2023-06-01 10:56:21 data-plane-kafka-request-handler-7-333 ERROR kafka.server.ControllerApis:76 - [ControllerApis nodeId=0] Unexpected error handling request Forwarded request: RequestContext(header=RequestHeader(apiKey=ENVELOPE, apiVersion=0, clientId=0, correlationId=2, headerVersion=2), connectionId='127.0.0.1:23621-127.0.0.1:58988-0', clientAddress=/127.0.0.1, principal=User:ANONYMOUS, listenerName=ListenerName(CONTROLLER), securityProtocol=PLAINTEXT, clientInformation=ClientInformation(softwareName=apache-kafka-java, softwareVersion=3.4.0), fromPrivilegedListener=false, principalSerde=Optional[org.apache.kafka.common.security.authenticator.DefaultKafkaPrincipalBuilder@4323763f]) RequestHeader(apiKey=DELETE_TOPICS, apiVersion=6, clientId=adminclient-12, correlationId=5, headerVersion=2) -- DeleteTopicsRequestData(topics=[DeleteTopicState(name='__org_kroxylicious_testing_consistencyTest', topicId=AAAAAAAAAAAAAAAAAAAAAA)], topicNames=[], timeoutMs=30000) with context RequestContext(header=RequestHeader(apiKey=DELETE_TOPICS, apiVersion=6, clientId=adminclient-12, correlationId=5, headerVersion=2), connectionId='127.0.0.1:23621-127.0.0.1:58988-0', clientAddress=/127.0.0.1, principal=User:ANONYMOUS, listenerName=ListenerName(CONTROLLER), securityProtocol=PLAINTEXT, clientInformation=ClientInformation(softwareName=unknown, softwareVersion=unknown), fromPrivilegedListener=false, principalSerde=Optional.empty)
org.apache.kafka.common.errors.TopicDeletionDisabledException: null
2023-06-01 10:56:21 awaitility-thread-386 WARN  io.kroxylicious.testing.kafka.common.Utils:171 - Failed to delete __org_kroxylicious_testing_consistencyTest. Caught: Topic deletion is disabled. 
org.apache.kafka.common.errors.TopicDeletionDisabledException: Topic deletion is disabled.
2023-06-01 10:56:21 EventHandler-360 INFO  kafka.log.UnifiedLog$:1787 - [LogLoader partition=__org_kroxylicious_testing_consistencyTest-0, dir=/var/folders/1g/2nsqtt916vg8f5bftn1tp0580000gn/T/kafka9657848742101429651/broker-0] Loading producer state till offset 0 with message format version 2
2023-06-01 10:56:21 EventHandler-360 INFO  kafka.log.LogManager:66 - Created log for partition __org_kroxylicious_testing_consistencyTest-0 in /var/folders/1g/2nsqtt916vg8f5bftn1tp0580000gn/T/kafka9657848742101429651/broker-0/__org_kroxylicious_testing_consistencyTest-0 with properties {}
2023-06-01 10:56:21 EventHandler-360 INFO  kafka.cluster.Partition:66 - [Partition __org_kroxylicious_testing_consistencyTest-0 broker=0] No checkpointed highwatermark is found for partition __org_kroxylicious_testing_consistencyTest-0
2023-06-01 10:56:21 EventHandler-360 INFO  kafka.cluster.Partition:66 - [Partition __org_kroxylicious_testing_consistencyTest-0 broker=0] Log loaded for partition __org_kroxylicious_testing_consistencyTest-0 with initial high watermark 0
2023-06-01 10:56:21 main-1 INFO  kafka.server.BrokerServer:66 - [BrokerServer id=0] Transition from STARTED to SHUTTING_DOWN
```

I can't see anything from `state.change.logger`

franvila: running again, I got three errors. This time the test was `TemplateTest$Tuples`. I don't get the same number of errors always, but there is no run without any of them 

```
Running io.kroxylicious.testing.kafka.junit5ext.TemplateTest$Tuples
Formatting /tmp/kafka8216214224463261426/broker-0 with metadata.version 3.3-IV0.
Formatting /tmp/kafka15363279691114438959/broker-0 with metadata.version 3.3-IV0.
Formatting /tmp/kafka15363279691114438959/broker-2 with metadata.version 3.3-IV0.
Formatting /tmp/kafka15363279691114438959/broker-1 with metadata.version 3.3-IV0.
2023-06-01 07:26:09 ERROR state.change.logger:74 - [Broker id=2] Received LeaderAndIsrRequest with correlation id 1 from controller 2 epoch 2 for partition __org_kroxylicious_testing_consistencyTest-0 (last update controller epoch 1) but cannot become follower since the new leader -1 is unavailable.
2023-06-01 07:26:09 ERROR state.change.logger:74 - [Broker id=0] Received LeaderAndIsrRequest with correlation id 1 from controller 2 epoch 2 for partition __org_kroxylicious_testing_consistencyTest-0 (last update controller epoch 1) but cannot become follower since the new leader -1 is unavailable.
2023-06-01 07:26:09 ERROR state.change.logger:74 - [Broker id=0] Received LeaderAndIsrRequest with correlation id 1 from controller 0 epoch 3 for partition __org_kroxylicious_testing_consistencyTest-0 (last update controller epoch 1) but cannot become follower since the new leader -1 is unavailable.
```

franvila: Agreed. I have already committed your changes

SamBarker (approved):
Thanks @franvila

k-wall:
Better late than never lgtm

PR #124, opened 2023-06-13
Create CHANGELOG.md

k-wall:


SamBarker (approved):
It would be good to have some tooling based on this which updates a releases page on kroxylicious.io and used to generate the GitHub release description too. 

This is a good start however.

PR #134, opened 2023-06-27
Support @BeforeEach and test cases referring to same cluster

robobario:
### Type of change

- Enhancement / new feature

### Description

Users can now refer to the same cluster from an `@BeforeEach` annotated method and the subsequent test method when using explicitly named clusters.

For example:

```java
@BeforeEach
public void setup(@Name("clusterA") KafkaCluster cluster) {
    .. do setup work ..
}

@Test
public void test(@Name("clusterA") KafkaCluster cluster) {
    .. do test work ..
}
```

Note that without the `@Name` annotations each parameter would have a unique cluster injected.

As a consequence it's now legal to have multiple references to the same cluster as long as later declarations only supply an `@Name` annotation and no constraint annotations (because re-declaring constraints for a cluster that already exists can't be done).

### legal example
```java
public void test(@BrokerCluster(numBrokers = 2) @Name("clusterA") KafkaCluster cluster, @Name("clusterA") KafkaCluster clusterRef) {
    .. do test work ..
}
```

### illegal examples
```java
public void test(@BrokerCluster(numBrokers = 2) @Name("clusterA") KafkaCluster cluster, @BrokerCluster(numBrokers = 1) @Name("clusterA") KafkaCluster clusterRef) {
    .. do test work ..
}
```

```java
// we disallow any constraints even if they're identical
public void test(@BrokerCluster(numBrokers = 2) @Name("clusterA") KafkaCluster cluster, @BrokerCluster(numBrokers = 2) @Name("clusterA") KafkaCluster clusterRef) {
    .. do test work ..
}
```

### Additional Context

Why:
We might want to extract some common setup to an `@BeforeEach` annotated method and have it refer to the same cluster as is used later in the test. For example to create some topics or assert some pre-conditions like cluster size for all cases.

Note this changes an existing behaviour where it was not allowed to have multiple KafkaClusters injected using the same name. This makes sense because a KafkaCluster declaration also implies the constraints required and the multiple declarations could have different constraints. So with the new change a parameter is allowed to refer to a cluster instantiated earlier, but not supply any constraints.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
@@ -792,10 +792,10 @@ public class KafkaClusterExtension implements
         if (sourceElement.isAnnotationPresent(Name.class)
                 && !sourceElement.getAnnotation(Name.class).value().isEmpty()) {
             clusterName = sourceElement.getAnnotation(Name.class).value();
-            Object o = store.get(clusterName);
-            if (o != null) {
+            if (store.get(clusterName) != null && !constraints.isEmpty()) {
                 throw new ExtensionConfigurationException(
-                        "A " + KafkaCluster.class.getSimpleName() + "-typed declaration with @Name(\"" + clusterName + "\") is already in scope");
+                        "A " + KafkaCluster.class.getSimpleName() + "-typed declaration with @Name(\"" + clusterName
+                                + "\") already exists, we cannot apply new constraints");
             }
         }
         else {
@@ -820,7 +820,6 @@ public class KafkaClusterExtension implements
                 extensionContext.getUniqueId(),
                 sourceElement,
                 clusterName);
-        cluster.start();
         return cluster;
     }
 
@@ -953,6 +952,7 @@ public class KafkaClusterExtension implements
                 extensionContext.getUniqueId(),
                 sourceElement,
                 clusterName);
+        c.start();
         return new Closeable<>(sourceElement, clusterName, c);
     }
 
--- a/junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
+++ b/junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
@@ -34,6 +34,7 @@ import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotSame;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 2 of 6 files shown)


robobario:
Planning to create a second PR which makes the @BeforeEach plus KafkaCluster combination not supported as I can see it could be seen as invalid to have multiple `KafkaCluster` declarations with this special meaning for a second declaration with no constraints (the empty set of constraints is still a constraint set).

Other ugly alternatives, have a flag or something on the `Name` annotation like `reuse=true` or add a new annotation like `@Reference` to capture the distinction.

SamBarker:
> Note this changes an existing behaviour where it was not allowed to have multiple KafkaClusters injected using the same name. This makes sense because a KafkaCluster declaration also implies the constraints required and the multiple declarations could have different constraints. So with the new change a parameter is allowed to refer to a cluster instantiated earlier, but not supply any constraints.

I think this is legitimate, injecting a cluster with the same name but different configuration seems like a bug to me, even if done intentionally it would make the tests very hard to reason about. So I'm fine with relaxing the restriction so that `@Name` can be a pointer as well as a restriction.

k-wall (approved):
looks reasonable to me.

robobario:
Oh and a note, it doesn't work with anonymous, un-named `KafkaCluster`. Every anonymous cluster declared is still unique. So you have to name your cluster for this `@BeforeEach` to refer to an existing cluster.

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
```diff
@@ -953,6 +952,12 @@ private static Closeable<KafkaCluster> createCluster(ExtensionContext extensionC
                 extensionContext.getUniqueId(),
                 sourceElement,
                 clusterName);
+        c.start();
+        LOGGER.log(TRACE,
```
SamBarker, line 956: Is it really worth including this?

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
```diff
@@ -111,6 +112,18 @@ public void twoDefinedClusterParameterAndAdmin(
         assertEquals(2, describeCluster(adminB).nodes().get().size());
     }
 
+    @Test
+    public void multipleReferencesToTheSameCluster(
+                                                   @BrokerCluster(numBrokers = 1) @Name("A") KafkaCluster clusterA,
+                                                   @Name("A") KafkaCluster clusterARef,
+                                                   @Name("A") Admin adminA)
+            throws ExecutionException, InterruptedException {
+        assertSameCluster(clusterA, adminA);
```
SamBarker, line 121: Should we also do this for `clusterA` and `clusterARef`?

SamBarker (approved):
I've raised https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/136 to address the repeatability of injecting anonymous clusters separately. 


Thanks @robobario

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
robobario: yeah, bit overkill, removed

PR #135, opened 2023-06-27
Prevent parameter injection into @BeforeEach/@AfterEach

robobario:
### Type of change

- Enhancement / new feature

### Description

Prevent injecting into BeforeEach and AfterEach annotated methods

For example:
```java
@ExtendWith(KafkaClusterExtension.class)
public class ExampleTest extends AbstractExtensionTest {


    @BeforeEach
    public void setup(KafkaCluster cluster){
        cluster.getBootstrapServers();
    }

    @Test
    public void tlsClusterParameter(KafkaCluster cluster) {
        fail("should never get here");
    }

}
```

will fail with:
```java
org.junit.jupiter.api.extension.ParameterResolutionException: Cannot inject interface io.kroxylicious.testing.kafka.api.KafkaCluster into method setup of class ExampleTest, incompatible with [@BeforeEach]
```

### Additional Context

Currently users can inject into methods annotated with BeforeEach and AfterEach, however the behaviour could be a bit surprising. If you inject KafkaCluster you will get a different cluster than the test method. To prevent these misunderstandings we could prevent users from injecting into BeforeEach/AfterEach methods.

This is an alternative to #134 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
@@ -5,8 +5,6 @@
  */
 package io.kroxylicious.testing.kafka.common;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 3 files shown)


SamBarker:
> If you inject KafkaCluster you will get a different cluster than the test method.

I consider that a bug. Injecting two nameless clusters into a single test method is just about Ok, personally still think they should be named, but I think it violates the principal of least surprise. Naively if I had a KafkaCluster annotation on both setup and test method I'd expect the same cluster.

robobario:
Preferring the solution in #134

PR #139, opened 2023-07-03
Fix BeforeEachTest so that it doesn't fail sporadically

k-wall:
### Type of change

- Bugfix

### Description

Fix `BeforeEachTest` so that it doesn't fail sporadically.

### Additional Context

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/BeforeEachTest.java
+++ b/junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/BeforeEachTest.java
@@ -5,20 +5,18 @@
  */
 package io.kroxylicious.testing.kafka.junit5ext;
 
-import java.util.Collection;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
-
+import io.kroxylicious.testing.kafka.api.KafkaCluster;
 import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.NewTopic;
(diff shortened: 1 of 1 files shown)


SamBarker (approved):
What was the underlying cause of the failure? As in why do we still have topics when the cluster is injected

robobario:
> What was the underlying cause of the failure? As in why do we still have topics when the cluster is injected

It's attempting to test that a topic created in the `BeforeEach` is available later in the test case (proving we are able to manipulate the same cluster from the `BeforeEach`. At a guess there must be some eventual consistency in there where the success of the topic creation doesn't guarantee it's available for listing. Maybe success means it was successfully written to the KRaft log but it's not been consumed back out yet.

robobario (approved):
LGTM

SamBarker:
> It's attempting to test that a topic created in the `BeforeEach` is available later in the test case (proving we are able to manipulate the same cluster from the `BeforeEach`. At a guess there must be some eventual consistency in there where the success of the topic creation doesn't guarantee it's available for listing. Maybe success means it was successfully written to the KRaft log but it's not been consumed back out yet.

Yeah `createTopic` returning just means the broker has the request not that the topic is available. See [Utils.awaitExpectedBrokerCountInClusterViaTopic](https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/12ee8a724451dbff65fbe1d18492cc1de704fe47/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java#L124-L125)

PR #142, opened 2023-07-07
Improve Utils#createTopic

k-wall:
### Type of change

_Select the type of your PR_

- Refactoring

### Description

- avoid ignoring the result of the recursive call
- avoid the tight spin that occurs whilst awaiting brokers to mesh etc (avoids screeds of output in the logs and the cpu spike)
- silences the `FutureReturnValueIgnored` complaint from sonatype-lift

### Additional Context

Separated from #138 to ease the review.

_Why are you making this pull request?_

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [x] Make sure all tests pass
- [ ] Update documentation
- [ ] Reference relevant issue(s) and close them after merging
- [x] Update CHANGELOG.md

PR #144, opened 2023-07-10
Fix regression caused by #142

k-wall:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

 Fix regression introduced into Utils#createTopic by #142 3cd6816bfa71b0961e6eb7d4616af8d038c4b0be

* Regression meant that the create topic looping was broken.  Exceptions occurring during stage handlers are propagated to subsequent handlers wrapped by a `CompletionException`.  The code forgot to handle this case.
* Use `failFast` to short circuit awaitExpectedBrokerCountInClusterViaTopic if the createTopic fails unexpectedly.  This will reveal the error that caused `createTopic` to us
* Also enabled archiving of the container logs in the CI workflow, in the case of failure.

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
@@ -121,10 +121,16 @@ public class Utils {
         try (var admin = Admin.create(connectionConfig)) {
             log.debug("Creating topic: {} via {}", CONSISTENCY_TEST, connectionConfig.get(BOOTSTRAP_SERVERS_CONFIG));
             // Note we don't wait for the Topic to be created, we assume it will complete eventually.
-            // As the thing we care about is it actually being replicated to the other brokers anyway so we wait to confirm replication.
-            var ignored = createTopic(expectedBrokerCount, admin);
+            // As the thing we care about is it actually being replicated to the other brokers anyway, so we wait to confirm replication.
+            var createTopicStage = createTopic(expectedBrokerCount, admin);
             log.debug("Waiting for {} to be replicated to {} brokers", CONSISTENCY_TEST, expectedBrokerCount);
             awaitCondition(timeout, timeUnit)
+                    .failFast(() -> {
+                        var f = createTopicStage.toCompletableFuture();
+                        if (f.isCompletedExceptionally()) {
+                            f.get();
+                        }
+                    })
                     .until(() -> {
                         log.debug("Calling describe topic");
                         final var promise = new CompletableFuture<Boolean>();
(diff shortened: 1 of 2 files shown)


k-wall:
The `failFast` will give us output like this.  Notice the **TerminalFailureException**

```
2023-07-10 14:24:46 WARN  io.kroxylicious.testing.kafka.common.Utils:218 - Failed to create topic: __org_kroxylicious_testing_consistencyTest due to java.lang.IllegalStateException: mocked

org.awaitility.core.TerminalFailureException: java.lang.RuntimeException: Failed to create topic: __org_kroxylicious_testing_consistencyTest

        at org.awaitility.core.ConditionAwaiter.executeFailFastConditionIfDefined(ConditionAwaiter.java:202)
        at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:97)
        at org.awaitility.core.CallableCondition.await(CallableCondition.java:78)
        at org.awaitility.core.CallableCondition.await(CallableCondition.java:26)
        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:985)
        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:954)
        at io.kroxylicious.testing.kafka.common.Utils.awaitExpectedBrokerCountInClusterViaTopic(Utils.java:135)
```

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -205,6 +211,9 @@ private static CompletionStage<Void> createTopic(Integer expectedBrokerCount, Ad
                 .thenRun(() -> log.debug("Create future for topic {} completed.", CONSISTENCY_TEST))
                 .exceptionallyComposeAsync((throwable) -> {
                     log.warn("Failed to create topic: {} due to {}", CONSISTENCY_TEST, throwable.getMessage());
+                    if (throwable instanceof CompletionException && throwable.getCause() != null) {
+                        throwable = throwable.getCause();
```
SamBarker, line 215: nit: I'm not a fan of re-assigning the parameter. I would prefer to see us delegate the handling of the throwable to a new method and call it with `throwable.getCause()` in this case and just `throwable` in others.

SamBarker (approved):
LGTM.

@k-wall address the nit or not as you deem appropriate but feel free to merge when your ready.

franvila (approved):

impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
k-wall: Ok, I see how an extract method looks.  I'll include it as part of #138 as that changes the method too.

PR #145, opened 2023-07-10
Direct storage formatting stdout to the log

k-wall:
### Type of change

- Enhancement / new feature

### Description

Kafka storage formatter is used by the in-VM path.  It currently logs to stdout.  It looks untidy and it means all the diagnostic output isn't in one place.  

### Additional Context

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
@@ -11,6 +11,13 @@ import java.io.OutputStream;
 import java.io.PrintStream;
 import java.nio.charset.Charset;
 
+/**
+ * Redirects a {@link PrintStream}'s print calls to the given {@link System.Logger} at
+ * a specified {@link java.lang.System.Logger.Level}. Line feed and/or carriage return characters are
+ * used to as delimiters, separating one log line from the next.
+ * <br/>
+ * The only use-case for this class is to catch a few lines of output from the {@link kafka.tools.StorageTool}.
+ */
 final class LoggingPrintStream {
 
     static PrintStream loggingPrintStream(System.Logger logger, System.Logger.Level level) {
@@ -32,6 +39,7 @@ final class LoggingPrintStream {
 
         @Override
         public void write(int b) {
+            // Note: the use-case of this class is so limited, a byte-by-byte approach is good enough.
             if (b == '\n' || b == '\r') {
                 logPending();
             }
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
@@ -46,13 +46,13 @@ final class LoggingPrintStream {
         }
 
         private void logPending() {
-            if(logger.isLoggable(level) {
-                 var log = os.toString(Charset.defaultCharset()).stripTrailing();
-                 if (!log.isEmpty()) {
-                     logger.log(level, "{0}", log);
-                 }
-                 os.reset();
+            if (logger.isLoggable(level)) {
+                var log = os.toString(Charset.defaultCharset()).stripTrailing();
+                if (!log.isEmpty()) {
+                    logger.log(level, "{0}", log);
+                }
             }
+            os.reset();
         }
     }
 
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright Kroxylicious Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.kroxylicious.testing.kafka.invm;
+
+import java.io.ByteArrayOutputStream;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.nio.charset.Charset;
+
+final class LoggingPrintStream {
+
+    static PrintStream loggingPrintStream(System.Logger logger, System.Logger.Level level) {
+        return new PrintStream(new LoggingOutputStream(logger, level));
+    }
+
+    private LoggingPrintStream() {
+    }
+
+    private static class LoggingOutputStream extends OutputStream {
+        private final ByteArrayOutputStream os = new ByteArrayOutputStream(1000);
+        private final System.Logger logger;
+        private final System.Logger.Level level;
+
+        public LoggingOutputStream(System.Logger logger, System.Logger.Level level) {
+            this.logger = logger;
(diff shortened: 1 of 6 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
```diff
@@ -0,0 +1,57 @@
...
+        public void close() {
+            logPending();
+        }
+
+        private void logPending() {
+            var log = os.toString(Charset.defaultCharset()).stripTrailing();
+            if (!log.isEmpty()) {
+                logger.log(level, "{0}", log);
+            }
+            os.reset();
```
SamBarker, line 53:
```suggestion
            if(logger.isLoggable(level) {
                 var log = os.toString(Charset.defaultCharset()).stripTrailing();
                 if (!log.isEmpty()) {
                     logger.log(level, "{0}", log);
                 }
                 os.reset();
            }
```

impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
```diff
@@ -0,0 +1,57 @@
...
+        private final System.Logger logger;
+        private final System.Logger.Level level;
+
+        public LoggingOutputStream(System.Logger logger, System.Logger.Level level) {
+            this.logger = logger;
+            this.level = level;
+        }
+
+        @Override
+        public void write(int b) {
```
SamBarker, line 34: IIRC for performance reasons its recommended to [override](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/OutputStream.html#:~:text=Subclasses%20are%20encouraged%20to%20override%20this%20method%20and%20provide%20a%20more%20efficient%20implementation.) `write(bytes, start, length)` do we care enough in context?

If not it might be worth a comment.

impl/src/test/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStreamTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Mock
+    private System.Logger logger;
+
+    @Test
+    void singleLog() {
+        try (var lps = LoggingPrintStream.loggingPrintStream(logger, System.Logger.Level.INFO)) {
+            lps.println("Hello, world");
+
+            verify(logger, times(1)).log(System.Logger.Level.INFO, "{0}", "Hello, world");
```
SamBarker, line 29: `times(1)` is implicit to `verify()` so arguably its just noise in the tests but they I wonder if it actually aids readability in context. The explicit call to times has me wondering if there should be confirming other numbers of invocations to.

impl/src/test/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStreamTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    void closeLogsUnfinishedPrint() {
+        try (var lps = LoggingPrintStream.loggingPrintStream(logger, System.Logger.Level.INFO)) {
+            lps.print("There's no newline");
+        }
+
+        verify(logger, times(1)).log(System.Logger.Level.INFO, "{0}", "There's no newline");
+        verifyNoMoreInteractions(logger);
+    }
```
SamBarker, line 65: style nit
```suggestion
    void closeLogsUnfinishedPrint() {
        // Given
        var lps = LoggingPrintStream.loggingPrintStream(logger, System.Logger.Level.INFO)) {
        lps.print("There's no newline");
        
        //When
        lps.close();
        
        //Then
        verify(logger, times(1)).log(System.Logger.Level.INFO, "{0}", "There's no newline");
        verifyNoMoreInteractions(logger);
    }
```

I think its worth making the call to close explicit as that is core to the test case.

impl/src/test/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStreamTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private System.Logger logger;
+
+    @Test
+    void singleLog() {
+        try (var lps = LoggingPrintStream.loggingPrintStream(logger, System.Logger.Level.INFO)) {
+            lps.println("Hello, world");
+
+            verify(logger, times(1)).log(System.Logger.Level.INFO, "{0}", "Hello, world");
+        }
+        verifyNoMoreInteractions(logger);
```
SamBarker, line 31: Given every single test case calls `verifyNoMoreInteractions(logger)` I wonder if it should really be pulled out to `afterEach`? 

Equally the warning from the mockito docs about [`verifyNoMoreInteractions`](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#finding_redundant_invocations) seems pertinent 
> A word of warning: Some users who did a lot of classic, expect-run-verify mocking tend to use verifyNoMoreInteractions() very often, even in every test method. verifyNoMoreInteractions() is not recommended to use in every test method. verifyNoMoreInteractions() is a handy assertion from the interaction testing toolkit. Use it only when it's relevant. Abusing it leads to overspecified, less maintainable tests.

So I'm kinda on the fence about it.

SamBarker (approved):
Hopefully this helps me figure out the broken pipe issues.

Thanks @k-wall

SamBarker:
Just wondering if you have come across [Streaming container logs](https://java.testcontainers.org/features/container_logs/#streaming-logs) from test containers?

impl/src/main/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStream.java
k-wall: Given this class has one tiny use-case (the storage formatter), I'm not sure it is worth the effort.  If i were going to apply extra effort any place it would be Kafka. I'd raise a KIP suggesting a less opinionated interface for kafka.tools.StorageTool#formatCommand.  PrintStreams are weird beasts.

I'll add a comment.

impl/src/test/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStreamTest.java
k-wall: I'll use the default.  I'd forgotten times(1) was implict.

impl/src/test/java/io/kroxylicious/testing/kafka/invm/LoggingPrintStreamTest.java
k-wall: My intent was to ensure that the class wasn't say, double logging, on close() or flush().  If I remove `verifyNoMoreInteractions`, then the test doesn't detect it.  I'll move verifyNoMoreInteractions to afterEach().  Unless there is a better way?

SamBarker: Nope I don't any better way!

PR #146, opened 2023-07-11
Make displayname available when using annotations

SamBarker:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Capture TestInfo during cluster creation from the test context so its available when using annotated parameters. This change has the unfortunate side effect of meaning the `testing-api` module now depends on `junit-api` (as we need the TestInfo in the `KafkaClusterProvisioningStrategy`) . Doing so with `provided` scope should limit the implications for downstream users.
 
### Additional Context

ClusterConfig exposes TestInfo which can be useful when generating names/labels for things especially in clusters. However the TestInfo was only set when users created the cluster manually rather than from annotations. 

The implementation is slightly clunky as junit doesn't provide an implementation of `TestInfo` in the public API so we need a record implementing the interface which is slightly annoying as `TestInfo` uses getter style and records use field style access...  

### Checklist

- [ ] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Please enumerate all user-facing changes using format `<githib issue/pr number>:
 
 ## 0.5.0
 
+* [#146](https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/146): Expose TestInfo (and thus DisplayName) when clusters are injected via  annotations.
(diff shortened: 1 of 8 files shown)


k-wall (approved):
lgtm

PR #147, opened 2023-07-13
Pummel podman

SamBarker:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Add an arbitrary cheap API call to try and ensure the container engine API is available.

### Additional Context

When running on podman on linux repeated executions of `TemplateTest` fail in the `Versions` test (this doesn't happen when run in isolation?!?!). This is currently a tactical hack to ensure the API is available. 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
@@ -51,6 +51,7 @@ import org.testcontainers.lifecycle.Startable;
 import org.testcontainers.lifecycle.Startables;
 import org.testcontainers.utility.DockerImageName;
 
+import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.command.InspectContainerCmd;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.exception.NotFoundException;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -577,29 +583,46 @@ private InspectContainerCmd buildInspectionCommandFor(KafkaContainer kc) {
         return kc.getDockerClient().inspectContainerCmd(kc.getContainerId());
     }
 
-    @SuppressWarnings({ "try" })
-    private static String createNamedVolume() {
-        var volumeCmd = DockerClientFactory.lazyClient().createVolumeCmd();
+    @SuppressWarnings({ "try", "resource" })
+    private static void createNamedVolume(String volumeName, String testName) {
+        pokeContainerEngine();
```
robobario, line 588: I think this logic should be put in front of the other `lazyClient` calls too as any of them could fail the same way if 5 seconds have elapsed and the podman process behind the systemd socket has stopped.

So maybe introduce a method so all the other usages would change from `DockerClientFactory.lazyClient` to:

```
@NotNull
    private static DockerClient getPreparedDockerClient() {
        DockerClient dockerClient = DockerClientFactory.lazyClient();
        pokeContainerEngine(dockerClient);
        return dockerClient;
    }
    ```

SamBarker: Yeah my intention was something very similar once I had validates this worked.

SamBarker:
I think this helped us un-cover the root cause but isn't in itself useful so I'm closing it.

PR #150, opened 2023-07-19
Use a new testcontainers DockerClient for every call

robobario:
### Type of change

- Bugfix

### Description

When running with podman we are struggling with the fact that the podman service by default kills idle connections after 10 seconds and our HTTP requests fail when trying to reuse pooled connections after they have idled more than 10 seconds. The request will be retried once by default so if there are 2+ connections that have expired server-side in the client pool both attempts fail.

This workaround creates a new client every time we do docker client work so that a new HttpClient will be created with a new pool. The code to load it is taken from [here](https://github.com/testcontainers/testcontainers-java/blob/1835f976c1b63c2be4c77b3e6b2c3802356db978/core/src/main/java/org/testcontainers/DockerClientFactory.java#L147).

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
@@ -690,6 +690,7 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
          */
         public OneShotContainer() {
             super(DockerImageName.parse("registry.access.redhat.com/ubi9/ubi-minimal"));
+            this.withStartupAttempts(CONTAINER_STARTUP_ATTEMPTS);
             this.withStartupCheckStrategy(new OneShotStartupCheckStrategy());
         }
     }
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
@@ -16,13 +16,16 @@ import java.nio.file.Path;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.OffsetDateTime;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -577,32 +581,44 @@ private InspectContainerCmd buildInspectionCommandFor(KafkaContainer kc) {
...
+            var volumeCmd = dockerClient.createVolumeCmd();
+            if (CONTAINER_ENGINE_PODMAN) {
+                volumeCmd.withDriverOpts(Map.of("o", "uid=" + KAFKA_CONTAINER_UID));
             }
+            var volumeName = volumeCmd.exec().getName();
+            volumesPendingCleanup.add(volumeName);
+            if (!CONTAINER_ENGINE_PODMAN) {
+                // On Docker, use a container to chown the volume.
+                // This is a workaround for https://github.com/moby/moby/issues/45714
+                try (var c = new OneShotContainer()) {
```
robobario, line 603: unfortunately this uses lazyClient internally in testcontainers code

robobario:
It's only a partial workaround, because all the container operations like start/stop use the global client under the hood they can suffer the same problem, it's just masked by container startup attempts.

k-wall (approved):

k-wall:
Thanks for the work @SamBarker and @robobario into this investigation.  I'm sure that's going to help us (and others) transition to Podman.

> It's only a partial workaround, because all the container operations like start/stop use the global client under the hood they can suffer the same problem, it's just masked by container startup attempts.

If this helps the situation on Linux, I'm +1 to merging.

FWIW I'm still seeing failures under podman on the Mac, even with this fix.    Sometimes rather than the ["failed to respond"](https://github.com/containers/podman/issues/17640)  from the Docker API call, I just see a hang.  When this happens for API calls other than create container,  the test times out before the 3 min default applied by the Docker client.     I think the root cause is still https://github.com/testcontainers/testcontainers-java/issues/7310 - it is just a different manifestation of the same problem (perhaps with a gvproxy dimension).    I've switched to Docker Desktop, so I'm not blocked.

Let's hope we can get a proper fix in the podman and/or docker-http client space so we can backout this change at some point.  Personally, I'm happy to live with the try-with-resources for the time being.

k-wall:
With service_timeout=0 fix I'm back to a green test suite, however, I'm still +1 on this change.  We may as well benefit from the robustness.

PR #151, opened 2023-07-19
Create DEV_GUIDE.md

k-wall:
### Type of change

_Select the type of your PR_

- Documentation

### Description

This project is short on development guidance. Let's fix that.

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/DEV_GUIDE.md
+++ b/DEV_GUIDE.md
@@ -1,0 +1,57 @@
+# Development Guide for Kroxylicious Junit5 Extension
+
+This document gives a detailed breakdown of the various build processes and options for building the Kroxylicious Junit5 Extension from source.
+
+## Build Pre-Requisites
+
+- [JDK](https://openjdk.org/projects/jdk/17/) (version 17 and above) - Maven CLI
+- [`mvn`](https://maven.apache.org/index.html) (version 3.5 and above) - Maven CLI
+- [`docker`](https://docs.docker.com/install/) or [`podman](https://podman.io/docs/installation) - Docker or Podman
+
+## JDK target
+
+The project targets language Java level 17.
+
+## Building / Running the Tests
+
+Build the project like this:
+
+```bash
+$ mvn clean install
+```
+
+The running of the tests can be controlled with the following Maven properties:
+
+| property           | description                                                                               |
+|--------------------|-------------------------------------------------------------------------------------------|
+| `-DskipUTs=true`   | skip unit tests                                                                           |
+| `-DskipITs=true`   | skip integration tests                                                                    |
+| `-DskipTests=true` | skip all tests                                                                            |
+| `-Pdebug`          | enables logging so you can see what the Kafka clients, Proxy and in VM brokers are up to. |
+
+The kafka environment used by the integrations tests can be _defaulted_ with these two environment variables.
+
+| env var                       | default | description                                                                                                                             |
+|-------------------------------|---------|-----------------------------------------------------------------------------------------------------------------------------------------|
+| `TEST_CLUSTER_EXECUTION_MODE` | `IN_VM` | `IN_VM` or `CONTAINER`. if `IN_VM`, kafka will be run same virtual machines as the integration test. Otherwise containers will be used. |
+| `TEST_CLUSTER_KRAFT_MODE`     | `true`  | if true, kafka will be run in kraft mode.                                                                                               |
+
+When the integration-tests are run in `CONTAINER` mode, the kafka/zookeeper logs are written to a location specified by
+the `container.logs.dir`  system property. When run through Maven this is defaulted to `integrationtests/target/container-logs`.
+
+Pass the `-Dquick` option to skip all tests and non-essential plug-ins and create the output artifact as quickly as possible:
+
+```bash
+$ mvn clean verify -Dquick
+```
+
+### Formatting
+
+Run the following command to format the source code and organize the imports as per the project's conventions:
+
+```bash
+$ mvn process-sources
+```
+
+
+


DEV_GUIDE.md
```diff
@@ -53,5 +55,48 @@ Run the following command to format the source code and organize the imports as
...
+```shell
+echo 'mkdir -p /etc/containers/containers.conf.d && printf "[engine]\nservice_timeout=0\n" > /etc/containers/containers.conf.d/service-timeout.conf' |  podman machine ssh --username root --
+```
+
+### Linux
+
+As a privileged user:
+
+```shell
+mkdir -p /etc/containers/containers.conf.d && printf "[engine]\nservice_timeout=0\n" > /etc/containers/containers.conf.d/service-timeout.conf
```
k-wall, line 82: Not verified.

robobario: On linux it might be better going into the user's home dir:
`mkdir -p  $HOME/.config/containers/ && printf "[engine]\nservice_timeout=0\n" > $HOME/.config/containers/containers.conf`

DEV_GUIDE.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+### Verify that the fix is effective
+
+Start this command:
+```shell
+socat - UNIX-CONNECT:/var/run/docker.sock
```
robobario, line 89: on my machine that's docker's sock (I have both installed) and `podman.sock` is at `/run/user/1000/podman/podman.sock`

DEV_GUIDE.md
robobario: confirming the /etc/ style works, let's go with that :+1:

SamBarker: `/etc` assumes rootful AFICT the `$HOME` version is used for rootless podman setups.

DEV_GUIDE.md
```diff
@@ -0,0 +1,102 @@
...
+
+- [JDK](https://openjdk.org/projects/jdk/17/) (version 17 and above) - Maven CLI
+- [`mvn`](https://maven.apache.org/index.html) (version 3.5 and above) - Maven CLI
+- [`docker`](https://docs.docker.com/install/) or [`podman`](https://podman.io/docs/installation) - Docker or Podman
+
+> :warning: **If you are using Podman please see the notes below**
+> 
+## JDK target
+
+The project targets language Java level 17.
```
SamBarker, line 15: Maybe better to say the project **production** sources target 17? As the tests depend on JDK 20 for SNI stuff.

DEV_GUIDE.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+### Code Formatting
+
+Run the following command to format the source code and organize the imports as per the project's conventions:
+
+```shell
+$ mvn process-sources
+```
+
```
SamBarker, line 57:
```suggestion
```

We suggest doing this before opening a Pull request as the build actions will fail the build for formatting which doesn't fit our conventions.
```

File: DEV_GUIDE.md [hunk 5]
```diff
@@ -0,0 +1,102 @@
...
+## Running Integration Tests on Podman
+
+There is an incompatibility between HTTP connection timeout expectations of 
+[testcontainers-java](https://github.com/testcontainers/testcontainers-java) and the Podman API. This
+can result in sporadic test failures when running the Integration Tests under Podman.  It manifests as
+failed or hanging REST API calls that lead to test failures and test hangs.
+
+It affects Linux and Mac OS X.
+
```
SamBarker, line 66: I think we need to give people some clear guidance around identifying the issue.
```suggestion
It affects Linux and Mac OS X.
On Linux it manifests as Http calls failing with a `Broken Pipe` exception. 
Similarly on MacOS we see a `localhost:XXX failed to respond`.
```

DEV_GUIDE.md
```diff
@@ -0,0 +1,102 @@
...
+To workaround around the issue, turn off the `service_timeout` by following these instructions.
+
+### MacOS X
+
+Start the `podman` machine as normal, then:
+
+```shell
+echo 'mkdir -p /etc/containers/containers.conf.d && printf "[engine]\nservice_timeout=0\n" > /etc/containers/containers.conf.d/service-timeout.conf' |  podman machine ssh --username root --
```
SamBarker, line 74: Setting `service_timeout=0` might be a bit extreme as it makes Podman a deamon process effectively which is not their intended usage model so maybe we should suggest 181 seconds to match the default HTTP timeout from Docker-java?

DEV_GUIDE.md
```diff
@@ -0,0 +1,102 @@
...
+### Code Formatting
+
+Run the following command to format the source code and organize the imports as per the project's conventions:
+
+```shell
+$ mvn process-sources
+```
+
+## Running Integration Tests on Podman
+
```
SamBarker, line 59:
````suggestion
It maybe necessary to configure the `DOCKER_HOST` environment variable to allow the tests to correctly use test containers.

```bash
DOCKER_HOST=unix://$( podman info --format '{{.Host.RemoteSocket.Path}}')
export DOCKER_HOST
```
````

SamBarker (approved):
Thanks @k-wall

DEV_GUIDE.md
k-wall: There's no SNI in the unit5 extension.  The whole project is Java 17.

DEV_GUIDE.md
k-wall: Oh, I've never needed that step. That's a Linux/podman thing, right?  Happy to add the suggestion.

DEV_GUIDE.md
k-wall: I think that's a good idea. I'll give that a test.  @robobario any thoughts?

DEV_GUIDE.md
k-wall: With podman machine, you get /var/run/docker.sock.
Guess I should use $(podman info --format '{{.Host.RemoteSocket.Path}}') in the verification command.

DEV_GUIDE.md
k-wall: Added as per suggestion.

DEV_GUIDE.md
SamBarker: Oh right, I'm think of the wrong project!

DEV_GUIDE.md
SamBarker: I'm not 100% sure when it's needed so I'd hesitate to say it's a Linux thing but yes that's where we have needed it.

DEV_GUIDE.md
robobario: Only that the idle timeout is 2x the number you specify. If you set --time=5 then the http idle is 10 seconds. Then 5 seconds after the last connection is closed, the process exits. So the process lasts 15 seconds if you make one request, and the connection is closed around 10 seconds. So you should be able to set it to 91.

DEV_GUIDE.md
robobario: On my machine with docker and podman both installed it sniffs that there is a file at `/var/run/docker.sock` and connects to that preferentially, which is docker for me. Not sure if that's true if you install podman only with no docker.

robobario: I need to set DOCKER_HOST to point it at the podman socket

SamBarker: Yeah thats probably the distinction, on my Mac `/var/run/docker.sock` points to Podman, but its missing on linux hence why I need to set DOCKER_HOST

DEV_GUIDE.md
k-wall: Yes, 91seconds works okay for me.

PR #152, opened 2023-07-19
Add a custom docker client strategy that configures the HTTP client f…

robobario:
…or podman

### Type of change

- Bugfix

### Description
Adds a custom docker client strategy that can be used with podman for resilience. This uses an http client with max 3 connections per route and a connection keepalive duration of 8 seconds. The connection pool will close them gracefully
if they have idle longer than that. Also as a failsafe I set retries to 4 so it could retry all connections for the route plus one in the worst case.

To enable this you must:
1. not have DOCKER_HOST set in your env
2. set TESTCONTAINERS_KROXY_DOCKER_HOST to point at your podman socket
3. set TESTCONTAINERS_DOCKER_CLIENT_STRATEGY to io.kroxylicious.testing.kafka.CustomStrategy

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/com/github/dockerjava/zerodep/TweakedDockerHttpClient.java
+++ b/impl/src/main/java/com/github/dockerjava/zerodep/TweakedDockerHttpClient.java
@@ -1,0 +1,282 @@
+/*
+ * Copyright Kroxylicious Authors.
+ *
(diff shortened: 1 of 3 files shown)


SamBarker:
@robobario is there still something we want to progress in this PR or is it dead?

robobario:
Let's close it, I don't think it's adding much.

PR #161, opened 2023-08-03
Trap system exit calls from the InVMKafkaCluster

SamBarker:
### Type of change

- Enhancement / new feature

### Description

This PR overrides the default SystemExit handler in Kafka to ensure the error is logged and a stack to trace it

### Additional Context

Kafka has a habit of committing seppuku when a variety of nasty things go wrong. This is *ahem* problematic in test suites. Luckily we can trap the call and suppress it to allow the tests to fail normally.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
@@ -30,6 +30,7 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import org.apache.kafka.common.utils.Exit;
 import org.apache.kafka.common.utils.Time;
 import org.apache.zookeeper.server.ServerCnxnFactory;
 import org.apache.zookeeper.server.ZooKeeperServer;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -427,4 +429,16 @@ public synchronized EndpointPair getEndpointPair(Listener listener, int nodeId)
         var port = portsAllocator.getPort(listener, nodeId);
         return EndpointPair.builder().bind(new Endpoint("0.0.0.0", port)).connect(new Endpoint("localhost", port)).build();
     }
+
+    private static void trapKafkaSystemExit() {
+        Exit.setExitProcedure((statusCode, message) -> {
+            final IllegalStateException illegalStateException = new IllegalStateException(message);
+            System.out.println("Kafka tried to exit with statusCode: " + statusCode + " and message: " + message);
```
SamBarker, line 436: System.out is not a long term solution but until we can confirm that the logger works its staying here.

robobario (approved):
LGTM

robobario:
Note there is a fix for the same thing in #161 along with trapping the `Exit` calls of Kafka. That fix looks good to me now while this set of futures I'm still thinking about how errors might propagate etc, so I think we should merge that PR and can do this refactor later.

gracegrimwood (approved):
Haven't managed to get this to produce a JVM crash error (as seen in #162), so LGTM 😃

k-wall (approved):

Issue #162, opened 2023-08-04
JVM intermittently crashes partway through tests


gracegrimwood:
**Describe the bug**
When running the kroxylicious-junit-extension tests, or when using the extension in other tests (specifically the Kroxylicious integration tests, see https://github.com/kroxylicious/kroxylicious/issues/508) the JVM intermittently crashes with something resembling the following error:
```java
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.1.2:test (default-test) on project testing-impl: 
[ERROR] 
[ERROR] Please refer to /Users/ggrimwoo/workspace/kroxylicious-junit5-extension/impl/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[ERROR] Command was /bin/sh -c cd '/Users/ggrimwoo/workspace/kroxylicious-junit5-extension/impl' && '/Users/ggrimwoo/.sdkman/candidates/java/20-tem/bin/java' '-jar' '/Users/ggrimwoo/workspace/kroxylicious-junit5-extension/impl/target/surefire/surefirebooter-20230803131446310_3.jar' '/Users/ggrimwoo/workspace/kroxylicious-junit5-extension/impl/target/surefire' '2023-08-03T13-14-43_682-jvmRun1' 'surefire-20230803131446310_1tmp' 'surefire_0-20230803131446310_2tmp'
[ERROR] Error occurred in starting fork, check output in log
[ERROR] Process Exit Code: 1
[ERROR] Crashed tests:
[ERROR] io.kroxylicious.testing.kafka.KafkaClusterTest
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
[...]

PR #163, opened 2023-08-07
Wait for kafka cluster shutdown before tempdir clean

robobario:
### Type of change

- Bugfix

### Description

When tearing down Kafka we now run through stages of teardown like:
1. shutdown kafka (brokers then controllers)
2. shutdown zookeeper
3. delete tempDir contents

Previously they could run simultaneously because the Kafka shutdown didn't await.

Hopefully this will prevent #162

### Additional Context

If kafka is running while the files are being deleted from under it this can result in Kafka halting/exiting the process as it detects horrifying action taking place. There is a separate PR to ensure a halt/exit doesn't kill tests and cause baffling test failures where the VM exits with code 1 and emits minimal logs.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
@@ -60,7 +60,8 @@ import static org.apache.kafka.server.common.MetadataVersion.MINIMUM_BOOTSTRAP_V
 public class InVMKafkaCluster implements KafkaCluster, KafkaClusterConfig.KafkaEndpoints {
     private static final System.Logger LOGGER = System.getLogger(InVMKafkaCluster.class.getName());
     private static final PrintStream LOGGING_PRINT_STREAM = LoggingPrintStream.loggingPrintStream(LOGGER, System.Logger.Level.DEBUG);
-    private static final int STARTUP_TIMEOUT = 30;
+    private static final int STARTUP_TIMEOUT_SECONDS = 30;
+    private static final int TEARDOWN_TIMEOUT_SECONDS = 30;
 
     private final KafkaClusterConfig clusterConfig;
     private final Path tempDirectory;
@@ -190,7 +191,7 @@ public class InVMKafkaCluster implements KafkaCluster, KafkaClusterConfig.KafkaE
     }
 
     private void tryToStartServerWithRetry(KafkaClusterConfig.ConfigHolder configHolder, Server server) {
-        Utils.awaitCondition(STARTUP_TIMEOUT, TimeUnit.SECONDS)
+        Utils.awaitCondition(STARTUP_TIMEOUT_SECONDS, TimeUnit.SECONDS)
                 .until(() -> {
                     // Hopefully we can remove this once a fix for https://issues.apache.org/jira/browse/KAFKA-14908 actually lands.
                     try {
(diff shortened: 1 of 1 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -328,7 +332,15 @@ public synchronized void stopNodes(Predicate<Integer> nodeIdPredicate, Terminati
                 .filter(e -> !stoppedServers.contains(e.getKey()))
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 
-        roleOrderedShutdown(kafkaServersToStop, true);
+        try {
+            roleOrderedShutdown(kafkaServersToStop).get(20, TimeUnit.SECONDS);
```
robobario, line 336: configurable timeout? or parameter in `stopNodes`?

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -352,49 +364,61 @@ public synchronized void startNodes(Predicate<Integer> nodeIdPredicate) {
...
+                                            .map(Path::toFile)) {
+                                s.forEach(File::delete);
+                            }
+                        }
+                    }
+                    catch (IOException e) {
+                        LOGGER.log(System.Logger.Level.ERROR, "error while deleting temp files", throwable);
+                        throw new RuntimeException(e);
+                    }
+                });
```
robobario, line 394: should we block on the cleanup? we could theoretically let it be cleaned in the background since each kafka/zk will use independent ports/tempdir. I think it would be cleaner to block.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
robobario: went with a hardcoded 30s matching the start timeout

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
robobario: added in a blocking get

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -352,49 +365,61 @@ public synchronized void startNodes(Predicate<Integer> nodeIdPredicate) {
...
-                        var s = ps
-                                .sorted(Comparator.reverseOrder())
-                                .map(Path::toFile)) {
-                    s.forEach(File::delete);
-                }
-            }
-        }
+        roleOrderedShutdown(servers)
+                .whenComplete((unused, throwable) -> {
+                    if (throwable != null) {
```
robobario, line 370: might be good to rethrow after each stage so an error bubbles out eventually to the `get`

robobario:
Note there is a fix for the same thing in #161 along with trapping the `Exit` calls of Kafka. That fix looks good to me now while this set of futures I'm still thinking about how errors might propagate etc, so I think we should merge that PR and can do this refactor later.

robobario:
Closing this, I think the complexity of adding futures to get to that slightly better high-level syntax isn't worth it at the moment

Issue #165, opened 2023-08-18
Unable to stand up a 3 controller KRaft-based cluster


tombentley:
**Describe the bug**

I'm unable to stand up a 3 controller KRaft-based cluster.

**To Reproduce**

```java

    @KRaftCluster(numControllers = 3)
    KafkaCluster cluster;

    @Test
    public void shouldWork() throws ExecutionException, InterruptedException {
        // Create some topics with partitions on known brokers
        var admin = Admin.create(Map.of(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, cluster.getBootstrapServers()));
        DescribeClusterResult describeClusterResult = admin.describeCluster();
        System.out.printf("Nodes %s%n", describeClusterResult.nodes().get());
        System.out.printf("Controller %s%n", describeClusterResult.controller().get());
    }
```

Declaring `cluster` as `TestcontianersKafkaCluster` does not make a difference.

**Expected behavior**

The cluster should be brought up and the test should pass after printing the nodes and controller

**Logs**

The test ultimately fails with this stacktrace:

```
org.awaitility.core.ConditionTimeoutException: Condition with io.kroxylicious.testing.kafka.invm.InVMKafkaCluster was not fulfilled within 30 seconds.

	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.CallableCondition.await(CallableCondition.java:78)
	at org.awaitility.core.CallableCondition.await(CallableCondition.java:26)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:985)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:954)
	at io.kroxylicious.testing.kafka.invm.InVMKafkaCluster.tryToStartServerWithRetry(InVMKafkaCluster.java:198)
	at io.kroxylicious.testing.kafka.invm.InVMKafkaCluster.lambda$start$5(InVMKafkaCluster.java:186)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:290)
[...]

PR #167, opened 2023-08-23
Support Multiple controllers

SamBarker:
### Type of change

- Bugfix
- Enhancement 

### Description

Fixes the cluster configuration and start up so that multiple controller nodes work. Fixes: https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/165

### Additional Context

I still see this as an intermediate step, there are kludges to detect if a node should be a controller or a broker or both. While they work it feels like we are missing a proper object model that knows what role a node has. We also lack the ability for users to control the cluster topology (i.e. they can't have nodes in controller only mode with while they they still have brokers to start)

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,17 @@ When multiple constraints are provided they will _all_ be satisfied.
 
 The cluster will be provisioned using the fastest available mechanism, because your development inner loop is a precious thing.
 
+## Provisioning topology
+
+In kraft cluster (the default) the extension will generate nodes using the `broker, controller` roles until it reaches the desired number of brokers or controllers (which ever is lowest) at which point it will continue deploying the remaining role.
+
+| numBrokers | numControllers | roles                                                               |
+|------------|----------------|---------------------------------------------------------------------|
+| 1          | 1              | `"broker,controller"`                                               |
+| 3          | 1              | `"broker,controller"`, `"broker"`, `"broker"`                       |
+| 1          | 3              | `"broker,controller"`, `"controller"`, `"controller"`               |
+| 3          | 3              | `"broker,controller"`, `"broker,controller"`, `"broker,controller"` |
+
 ## Provisioning mechanisms
 
 The following provisioning mechanisms are currently supported:
@@ -231,6 +242,3 @@ See the [developer guide](DEV_GUIDE.md).
 ## Releasing this project
 
 See the [releasing guide](RELEASING.md).
-
-
-
--- a/README.md
+++ b/README.md
@@ -74,9 +74,10 @@ When multiple constraints are provided they will _all_ be satisfied.
 
 The cluster will be provisioned using the fastest available mechanism, because your development inner loop is a precious thing.
 
-## Provisioning topology
+## Node topology
 
-In a kraft cluster (the default) the extension will generate nodes using the `broker, controller` roles until it reaches the desired number of brokers or controllers (which ever is lowest) at which point it will continue deploying the remaining role.
+When generating a cluster using KRaft (the default), you declare how many brokers and controllers you want and the extension will provision the minimum number of nodes to satisfy those conditions. It will create as many nodes as it can that are both KRaft controllers and brokers using [process.roles](https://kafka.apache.org/documentation/#brokerconfigs_process.roles) (process.roles = "broker,controller"). For example:
+
 
 | numBrokers | numControllers | roles                                                               |
 |------------|----------------|---------------------------------------------------------------------|
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ The cluster will be provisioned using the fastest available mechanism, because y
 
 ## Provisioning topology
 
-In kraft cluster (the default) the extension will generate nodes using the `broker, controller` roles until it reaches the desired number of brokers or controllers (which ever is lowest) at which point it will continue deploying the remaining role.
+In a kraft cluster (the default) the extension will generate nodes using the `broker, controller` roles until it reaches the desired number of brokers or controllers (which ever is lowest) at which point it will continue deploying the remaining role.
 
 | numBrokers | numControllers | roles                                                               |
 |------------|----------------|---------------------------------------------------------------------|
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -302,13 +302,19 @@ public class KafkaClusterConfig {
     private String determineRole(int nodeId) {
         var roles = new ArrayList<String>();
 
-        if (nodeId < brokersNum) {
+        if (nodeId < brokersNum || isAdditionalNode(nodeId)) {
             roles.add(BROKER_ROLE);
         }
         if (nodeId < kraftControllers) {
             roles.add(CONTROLLER_ROLE);
         }
         return String.join(",", roles);
+    }
+
+    // additional nodes can only be added after the initial topology is generated.
+    // Hence, it is safe to assume that a node is additional if it has a higherId than the initial topology would allow for.
+    private boolean isAdditionalNode(int nodeId) {
+        return nodeId >= Math.max(brokersNum, kraftControllers);
     }
 
     private void configureTls(KafkaEndpoints.EndpointPair clientEndpoint, Properties server) {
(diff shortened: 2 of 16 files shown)


impl/src/test/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfigTest.java
```diff
@@ -39,13 +41,89 @@ void generateConfigForSpecificBroker() {
...
+        // When
+        final Stream<KafkaClusterConfig.ConfigHolder> brokerConfigs = kafkaClusterConfig.getBrokerConfigs(() -> endpointConfig);
+
+        // Then
+        assertThat(brokerConfigs).hasSize(3);
+    }
+
+    private void assertNodeIdHasRole(KafkaClusterConfig kafkaClusterConfig, int nodeId, String expectedRole) {
+        final var config = kafkaClusterConfig.generateConfigForSpecificNode(endpointConfig, nodeId);
+        assertThat(config.getProperties()).extracting(brokerConfig -> brokerConfig.get("process.roles")).as("nodeId: %s to have process.roles", nodeId).isEqualTo(
```
robobario, line 123: We should switch to using the kafka enums throughout in another PR.

SamBarker: Ohh that is a good thought

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactory.java
```diff
@@ -74,6 +73,12 @@ public static KafkaCluster create(KafkaClusterConfig clusterConfig) {
             builder.kraftMode(kraftMode);
         }
 
+        if (KafkaClusterExecutionMode.CONTAINER == clusterMode && kraftMode && clusterConfig.getBrokersNum() < clusterConfig.getKraftControllers()) {
+            throw new IllegalStateException(
+                    "Due to https://github.com/ozangunalp/kafka-native/issues/88 we can't support controller only nodes so we need to fail fast. This cluster has "
```
robobario, line 78: maybe this log should mention the execution mode, the other one doesn't need to since the classname contains TestContainers

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -210,92 +215,95 @@ public Stream<ConfigHolder> getBrokerConfigs(Supplier<KafkaEndpoints> endPointCo
...
-        listeners.put("ANON", anonEndpoint.listenAddress());
-        advertisedListeners.put("ANON", anonEndpoint.advertisedAddress());
-
-        protocolMap.put("INTERNAL", SecurityProtocol.PLAINTEXT.name());
-        listeners.put("INTERNAL", interBrokerEndpoint.listenAddress());
-        advertisedListeners.put("INTERNAL", interBrokerEndpoint.advertisedAddress());
-        earlyStart.add("INTERNAL");
-        putConfig(server, "inter.broker.listener.name", "INTERNAL");
+        final ConfigHolder configHolder;
+        if (role.contains(BROKER_ROLE)) {
```
robobario, line 230: Worth pulling out a `brokerConfigHolder(...)` method and a `controllerOnlyConfigHolder(...)` type method? This block is big.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactory.java
```diff
@@ -93,6 +98,11 @@ public static KafkaCluster create(KafkaClusterConfig clusterConfig) {
         }
     }
 
+    private static KafkaClusterExecutionMode getExecutionMode(KafkaClusterConfig clusterConfig) {
+        return KafkaClusterExecutionMode.convertClusterExecutionMode(System.getenv().get(TEST_CLUSTER_EXECUTION_MODE),
```
robobario, line 102: not part of your changes but it looks a bit dangerous that it defaults it if the string is unknown, maybe better to explode with error if the user supplies value "CONTAINDERER"

robobario (approved):
LGTM, good to have all the language aligned on node/broker/controller thanks @SamBarker 

Would be good to have another round on `KafkaClusterConfig` to try and move away from going from the configuration straight to server properties, passing those maps down to collect protocols/listeners/etc. Maybe there's some intermediate representation missing like: `TestDefinition -> Nodes -> Properties` where `Nodes` describes how we want the cluster to look.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -216,7 +216,7 @@ private KafkaContainer buildKafkaContainer(KafkaClusterConfig.ConfigHolder holde
                 // KAFKA_LOG_DIR overrides a key in the quarkus kafka image application.properties. The quarkus app uses
                 // that to set log.dir. Any value we set for log.dir in server.properties is lost.
                 .withEnv("KAFKA_LOG_DIR", getBrokerLogDirectory(holder.getBrokerNum()))
-                .withEnv("QUARKUS_LOG_LEVEL", "DEBUG") // Enables org.apache.kafka logging too
+//                .withEnv("QUARKUS_LOG_LEVEL", "DEBUG") // Enables org.apache.kafka logging too
```
tombentley: Revert or remove?

tombentley:
We should probably update the README and the javadocs on the `@KRaftCluster` annotation to explain how the `numBrokers` and `numControllers` relate to the cluster that gets created. 

I also think that with this change you're not far from being able to add support for controlling whether we use combined or pure-controller nodes using some `combinedMode` attribute on `@KRaftCluster`. Maybe that's something we can do in a follow-up PR, WDYT?

SamBarker:
> We should probably update the README and the javadocs on the `@KRaftCluster` annotation to explain how the `numBrokers` and `numControllers` relate to the cluster that gets created.

Yeah thats a good point.

> I also think that with this change you're not far from being able to add support for controlling whether we use combined or pure-controller nodes using some `combinedMode` attribute on `@KRaftCluster`. Maybe that's something we can do in a follow-up PR, WDYT?

Yeah I've questioned the use of combined mode throughout this work but I really don't want to try and tackle it as part of this PR. Partly because I wonder if at that point we want to move to a more sophisticated deployment model such as `@KraftCluster(brokerIds = [1,2,4], controllerId=1,2,6])` but keep `numBrokers`/ `numControllers` as a fall back...

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
SamBarker: Just to make sure as I think there may have been a push while the review was in flight. Commenting this out aligns in with `main` is that what you were looking for or were you asking for the debug logging to be re-enabled?

tombentley: Yeah, I think this one is stale. I was just wanting to avoid having commented out code.

SamBarker: Ok will. remove the commented out line from here and the ZK container

SamBarker: fixed with https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/167/commits/8be2a6cf2ac5b17ed3d10598019f30cfa573fb96

README.md
```diff
@@ -74,6 +74,17 @@ When multiple constraints are provided they will _all_ be satisfied.
 
 The cluster will be provisioned using the fastest available mechanism, because your development inner loop is a precious thing.
 
+## Provisioning topology
+
+In kraft cluster (the default) the extension will generate nodes using the `broker, controller` roles until it reaches the desired number of brokers or controllers (which ever is lowest) at which point it will continue deploying the remaining role.
```
robobario, line 79: > In kraft cluster

Should we introduce some term like cluster mode [kraft, zookeeper]?

SamBarker: Hmm,  my first draft locally I called it `kraft mode` but thought `a kraft cluster` aligned better with the use of the `KraftCluster` annotation. 

The mode terminology would be extension specific as Apache Kafka talks about KRaft clusters IIRC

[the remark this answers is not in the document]
SamBarker: Certainly reads better to me

robobario (approved):
LGTM

README.md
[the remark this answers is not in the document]
SamBarker: resolved with https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/167/commits/d3ac926c0c999bc89b2980be7ddfb45ea52086dc

PR #177, opened 2023-09-12
Fix #176: Changes needed to get the tests working, at least

### LINKED ISSUE #176: Support both KRaft combined mode and separate controllers
Currently the extension supports ZooKeeper mode (using `@ZooKeeperCluster` as the constraint annotation) and KRaft in combined mode (using `@KraftCluster`). It would be good to also support KRaft using separate controllers.

This could be expressed in the API in several ways. Perhaps the simplest is just to add a `combinedMode` attribute to `@KraftCluster`. 

Although the API is simple the knock-on consequences are not. 

* `KafkaClusterConfig`  has a `kraftMode` boolean flag that is perhaps better treated as a 3 element `MetadataMode` enum (of ZK, Kraft combined and Kraft separate). 
* `KafkaClusterConfig`  has a `getBrokerConfigs()` that would now be more clearly called `getNodeConfigs()`, since not all nodes are brokers.
* `ConfigHolder` has `getBrokerNum()` that would now be more clearly called `getNodeId()`, again because not all nodes are brokers. 
* `KafkaCluster.getNumOfBrokers()` could remain a count of the brokers (not of all the nodes), but it might also be useful to have a count of all nodes. 
* The test coverage, which already has to cope with (ZK, combined) × (in-VM, containers)  would need to also cover the separate controller cases.

See discussion in #174 which led to this issue.

tombentley:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Adds support for running Kraft-based clusters with separate controller nodes.  

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KRaftCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KRaftCluster.java
@@ -17,25 +17,42 @@ import io.kroxylicious.testing.kafka.api.KafkaClusterProvisioningStrategy;
 /**
  * Annotation constraining a {@link KafkaClusterProvisioningStrategy} to use
  * a {@link KafkaCluster} that is KRaft-based.
+ *
+ * <table>
+ *  <caption>Breakdown of the interaction between numControllers and numBrokers</caption>
+ *  <tr><th>numBrokers</th><th>numControllers</th><th>combinedMode</th><th>roles</th></tr>
+ *  <tr><td>1</td><td>1</td><td>true</td><td>1×<code>"broker,controller"</code></td></tr>
+ *  <tr><td>1</td><td>1</td><td>false</td><td>1×<code>"broker"</code>, 1×<code>"controller"</code></td></tr>
+ *
+ *  <tr><td>3</td><td>1</td><td>true</td><td>1×<code>"broker,controller"</code>, 2×<code>"broker"</code></td></tr>
+ *  <tr><td>3</td><td>1</td><td>false</td><td>3×<code>"broker"</code>, 1×<code>"controller"</code>, </td></tr>
+ *
+ *  <tr><td>1</td><td>3</td><td>true</td><td>1×<code>"broker,controller"</code>, 2×<code>"controller"</code></td></tr>
+ *  <tr><td>1</td><td>3</td><td>false</td><td>1×<code>"broker"</code>, 3×<code>"controller"</code></td></tr>
+ *
+ *  <tr><td>3</td><td>3</td><td>true</td><td>3×<code>"broker,controller"</code></td></tr>
+ *  <tr><td>3</td><td>3</td><td>true</td><td>3×<code>"broker"</code>, 3×<code>"controller"</code></td></tr>
+ * </table>
  */
 @Target({ ElementType.PARAMETER, ElementType.FIELD })
 @Retention(RetentionPolicy.RUNTIME)
 @KafkaClusterConstraint
 public @interface KRaftCluster {
     /**
-     * The number of kraft controllers
-     * The extension will ensure there are enough nodes started with the <code>controller</code> role.
-     * The extension will combine this with the <code>numBrokers</code> to generate a cluster topology.
-     * <table>
-     *  <caption>Breakdown of the interaction between numControllers and numBrokers</caption>
-     *  <tr><th>numBrokers</th><th>numControllers</th><th>roles</th></tr>
-     *  <tr><td>1</td><td>1</td><td><code>"broker,controller"</code></td></tr>
-     *  <tr><td>3</td><td>1</td><td><code>"broker,controller"</code>, <code>"broker"</code>, <code>"broker"</code></td></tr>
-     *  <tr><td>1</td><td>3</td><td><code>"broker,controller"</code>, <code>"controller"</code>, <code>"controller"</code></td></tr>
-     *  <tr><td>3</td><td>3</td><td><code>"broker,controller"</code>, <code>"broker,controller"</code>, <code>"broker,controller"</code></td></tr>
-     * </table>
+     * The number of kraft controllers.
+     * The extension will ensure there are this many nodes started with the <code>controller</code> role.
+     * combining this with the <code>numBrokers</code> and <code>combinedMode</code> to generate a cluster topology.
+     *
+     * See the class JavaDoc for example topologies.
      * @return The number of KRaft controllers
      */
     public int numControllers() default 1;
+
+    /**
+     * Whether to use combined mode, where controllers can share a JVM with brokers.
+     * See the class JavaDoc for example topologies.
+     * @return true to use combined mode.
+     */
+    public boolean combinedMode() default true;
 
 }
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -216,7 +216,6 @@ public class KafkaClusterConfig {
      */
     @NotNull
     public ConfigHolder generateConfigForSpecificNode(KafkaEndpoints kafkaEndpoints, int nodeId) {
-        // checkNodeId(nodeId);
         final var roles = nodeId >= numNodes() ? EnumSet.of(NodeRole.BROKER) : processRoles(nodeId);
         Properties nodeConfiguration = new Properties();
         nodeConfiguration.putAll(brokerConfigs);
@@ -267,7 +266,7 @@ public class KafkaClusterConfig {
             configureKraftNode(kafkaEndpoints, nodeId, nodeConfiguration, protocolMap, listeners, earlyStart, roles);
         }
         else if (this.metadataMode == MetadataMode.ZOOKEEPER) {
-            configureLegacyBroker(nodeId, kafkaEndpoints, nodeConfiguration);
+            configureLegacyBroker(kafkaEndpoints, nodeConfiguration);
         }
 
         putConfig(nodeConfiguration, "listener.security.protocol.map",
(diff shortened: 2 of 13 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/common/KRaftCluster.java
```diff
@@ -17,25 +17,42 @@
...
      * @return The number of KRaft controllers
      */
     public int numControllers() default 1;
 
+    /**
+     * Whether to use combined mode, where controllers can share a JVM with brokers.
+     * See the class JavaDoc for example topologies.
+     * @return true to use combined mode.
+     */
+    public boolean combinedMode() default true;
```
SamBarker, line 56: Does this really make sense? I admit its what we talked about in the other PR and thus what you've done. 

I wonder if we should just expose `MetadataMode` here instead? I guess `metadataMode = Zookeeper` doesn't make sense here.

I almost wonder if we should drop the `KraftCluster` & `ZookeeperCluster` annotations and just have `KafkaCluster(medataData = [KRaftCombined | KRaftIndependent | Zookeeper]` ?

tombentley: I agree it's a bit murky. It is certainly true that we could express all topologies using something like this:

```java
@KafkaCluster(numXxx=X, metadataMode=Y, numBrokers=Z)
```

But I'm not sure what to call `numXxx`. It can't be `numControllers`, because in ZK mode the controller is a broker and the ZK nodes are counted separately. Maybe `metadataQuorumSize` would work?

I'm inclined to do the simplest possible thing in this PR and consider the API question separately. If nothing else to make the PRs of a sensible size.

SamBarker: Yeah I think separate PR for an api change makes sense. 

I'll give this another pass with that in mind.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -297,16 +313,16 @@ public synchronized int addBroker() {
...
     @Override
     public synchronized void removeBroker(int nodeId) throws IllegalArgumentException, UnsupportedOperationException {
         if (!servers.containsKey(nodeId)) {
-            throw new IllegalArgumentException("Broker node " + nodeId + " is not a member of the cluster.");
+            throw new IllegalArgumentException("Node " + nodeId + " is not a member of the cluster.");
         }
-        if (clusterConfig.isKraftMode() && isController(nodeId)) {
-            throw new UnsupportedOperationException("Cannot remove controller node " + nodeId + " from a kraft cluster.");
+        if (!clusterConfig.isPureBroker(nodeId)) {
+            throw new UnsupportedOperationException("Node " + nodeId + " is not a pure broker.");
```
SamBarker, line 325: I understand what this message means but is the `pure broker` terminology going to be meaningful to users?

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -315,7 +331,7 @@ public synchronized void removeBroker(int nodeId) throws IllegalArgumentExceptio
             throw new IllegalStateException("Cannot remove nodes from a cluster with stopped nodes.");
         }
 
-        var target = servers.keySet().stream().filter(n -> n != nodeId).findFirst();
+        var target = servers.keySet().stream().filter(n -> n != nodeId && clusterConfig.hasBrokerRole(n)).findFirst();
```
SamBarker, line 334: is `&& clusterConfig.hasBrokerRole(n)` really needed in the filter clause? Shouldn't we have validated that up front that `clusterConfig.hasBrokerRole(nodeId)` is true?

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -396,15 +412,13 @@ private void roleOrderedShutdown(Map<Integer, Server> servers) {
     }
 
     private boolean isController(Integer key) {
-        // TODO this is nasty. We shouldn't need to go via the portAllocator to figure out what a node is
-        // But it is at least testing something meaningful about the configuration
-        return portsAllocator.hasRegisteredPort(Listener.CONTROLLER, key);
+        return key < clusterConfig.numNodes() // dynamically added nodes are always pure brokers
+                && clusterConfig.hasControllerRole(key);
```
SamBarker, line 416: I think pushing this back to the `clusterConfig` is a good move, but it  moves us back to playing games with nodeId's rather testing for something that is an inherent property of the feature we care about.

Separately I don't see why we need the first half of the condition `key < clusterConfig.numNodes()` it either has the controller role or not. Sprinkling `key < clusterConfig.numNodes()` throughout just defuses the definition of what is and isn't a controller throughout the codebase.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -396,15 +412,13 @@ private void roleOrderedShutdown(Map<Integer, Server> servers) {
...
     }
 
     private boolean isBroker(Integer key) {
-        // TODO this is nasty. We shouldn't need to go via the portAllocator to figure out what a node is
-        // But it is at least testing something meaningful about the configuration
-        return portsAllocator.hasRegisteredPort(Listener.ANON, key);
+        return key >= clusterConfig.numNodes() // dynamically added nodes are always pure brokers
```
SamBarker, line 420: For the record similar thoughts apply here as to `isController` above

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -185,15 +187,26 @@ public TestcontainersKafkaCluster(DockerImageName kafkaImage, DockerImageName zo
...
-            portAllocationSession.allocate(Set.of(Listener.CONTROLLER), 0, clusterConfig.getKraftControllers());
+            for (int nodeId = 0; nodeId < clusterConfig.numNodes(); nodeId++) {
+                Set<Listener> listeners = new HashSet<>();
+                if (clusterConfig.hasControllerRole(nodeId)) {
+                    listeners.add(Listener.CONTROLLER);
+                }
+                if (clusterConfig.hasBrokerRole(nodeId)) {
+                    listeners.add(Listener.EXTERNAL);
+                    listeners.add(Listener.ANON);
+                    listeners.add(Listener.INTERNAL);
```
SamBarker, line 198: Do we actually need to expose `INTERNAL` outside of the docker/podman network?

I didn't think we did which is why it wasn't previously included in the port allocation

SamBarker (approved):
I'm happy with the general shape of this. A few thoughts to be discussed about the details.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
tombentley: cluster config is immutable and encodes the requested number of brokers and controllers. When nodes are added dynamically it doesn't know about them. We might be able to have it reason about `node.id > numNodes()`, but when what does `numNodes()` actually mean? Since it's the `KafkaCluster` that is exposing `add/removeBroker()` it seems reasonable to expect it to be responsible for knowing about the node ids required.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
tombentley: This is about finding the target brokers for replicas currently on `nodeId`, I.e. `n` here is not the same as `nodeId`, and without the extra condition `n` might be a controller, but not a broker, thus cannot replicate any partitions hosted the broker which is being removed.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
SamBarker: I'm not sure if we are agreeing or not. 

I think guessing if something is a controller or not based on its ID is not a good idea. It just bakes in assumptions about the order of operations. The old model of checking the port allocations was an odd choice but at least tested for a property of being a controller or not.

It still strikes me that we should have some way to say given a nodeID what roles does it have. Pushing the question of what does this nodeID do into clusterConfig and thus hiding the nodeID guessing is a decent compromise.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -230,18 +228,45 @@ public ConfigHolder generateConfigForSpecificNode(KafkaEndpoints kafkaEndpoints,
         var earlyStart = new TreeSet<String>();
 
         final ConfigHolder configHolder;
-        if (role.contains(BROKER_ROLE)) {
-            configHolder = configureBroker(kafkaEndpoints, nodeId, protocolMap, listeners, advertisedListeners, earlyStart, nodeConfiguration);
+        if (NodeRole.hasBrokerRole(roles)) {
+            var interBrokerEndpoint = kafkaEndpoints.getEndpointPair(Listener.INTERNAL, nodeId);
```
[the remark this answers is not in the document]
tombentley: Broker and controller are non-exclusive roles which a node may have. Having separate methods for broker and controller obscures the combined case. Certainly we can try to tidy this up a bit, but I think it would be better to move the ConfigHolder instantiation outside of the if/else, to make it clear that it's only the endpoints which are different when a node is a broker.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
tombentley: I _think_ we agree? With what I've done in this PR then given a `MetadataMode` and the number of brokers and controllers it's completely deterministic what roles any given node id should have. There is no guessing, and the order of operations cannot change the role:  If a node ends up configured with different roles it's simply a bug.

k-wall:
I'm going to guess this is stale.  We have rudimentary support for forming KRaft clusters with nodes in the controller only role (which seems to be passing the good enough test).

PR #180, opened 2023-09-19
Update kafka version range in testVersions to include 3.5.1

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

`kafka-native` has just published Kafka 3.5.1.  The Junit extension starts to use this by default, however, the enumeration of versions tested by testVersions should be extended.

### Additional Context

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please enumerate all user-facing changes using format `<githib issue/pr number>:
 
 ## 1.0.0
 
+## 0.6.0
+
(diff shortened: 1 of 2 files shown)


SamBarker (approved):
Good catch @k-wall we should maybe revisit the test so new releases don't just break the build

franvila (approved):
Good catch!

PR #184, opened 2023-09-21
fix: stop Utils#createTopic trying to create topic if admin client gets closed

k-wall:
### Type of change

- Bugfix

### Description

Break out of the createTopic creation loop if the `RetriableException` message indicates that the Admin client is closed.  This prevents a thread pointlessly spinning in the background, failing forever.

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
@@ -227,7 +227,10 @@ public class Utils {
         // If the throwable originates from within a CompletionStage's handler, it will be wrapped within a
         // CompletionException.
         var throwable = potentiallyWrapped instanceof CompletionException && potentiallyWrapped.getCause() != null ? potentiallyWrapped.getCause() : potentiallyWrapped;
-        return throwable instanceof RetriableException || throwable instanceof InvalidReplicationFactorException
+        boolean retriable = throwable instanceof RetriableException
+                && (throwable.getMessage() == null
+                        || !throwable.getMessage().contains("The AdminClient is not accepting new calls") /* workaround for KAFKA-15507 */ );
+        return retriable || throwable instanceof InvalidReplicationFactorException
                 || (throwable instanceof TopicExistsException && throwable.getMessage().contains("is marked for deletion"));
     }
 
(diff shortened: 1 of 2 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/common/Utils.java
```diff
@@ -227,7 +227,9 @@ private static boolean isRetryable(Throwable potentiallyWrapped) {
         // If the throwable originates from within a CompletionStage's handler, it will be wrapped within a
         // CompletionException.
         var throwable = potentiallyWrapped instanceof CompletionException && potentiallyWrapped.getCause() != null ? potentiallyWrapped.getCause() : potentiallyWrapped;
-        return throwable instanceof RetriableException || throwable instanceof InvalidReplicationFactorException
+        boolean retriable = throwable instanceof RetriableException
+                && (throwable.getMessage() == null || !throwable.getMessage().contains("The AdminClient is not accepting new calls"));
```
k-wall, line 231: I don't see a nicer way to actually discover that the AdminClient is closed.

SamBarker: I had wondered if we should just throw away the admin client and create a new one...

k-wall: No, It is the caller's responsibility to pass the Admin, so I don't think it would be appropriate for createTopic to spontaneously create a new one. 

The caller is awaitExpectedBrokerCountInClusterViaTopic.  I think we can infer that awaitExpectedBrokerCountInClusterViaTopic has timed-out, and the try-with-resources has closed the Admin client.    In the case where Admin client becomes close, we want createTopic to give up.

k-wall:
@SamBarker what do you think to committing this change?  I think in the long term, the Kafka Client needs a fix, but this would improve the behaviour in the meanwhile.

SamBarker (approved):
Its nasty but if the Admin client is going to lie to us about the type of error then I don't see we can do much else.

k-wall:
I'll comment the nastiness with a note about https://issues.apache.org/jira/browse/KAFKA-15507. Hopefully we can kill the code one day.

PR #187, opened 2023-09-29
Reduce the scope of fields the extension will inject into.

SamBarker:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

This change restricts the extension to only injecting into fields where there are no annotations, or only annotations known to be safe which are those coming from:
- `io.kroxylicious` 
- `org.junit`
- `java.lang`

It also prevents injection into fields which already have a non null value.

### Additional Context

I've had a couple of issues with the extension trying to inject into fields without having a KafkaCluster field and thus the injection fails. In my specific tests I didn't want the extension to inject into the fields. As I was using a mockito to mock the admin client or building my own `MockProducer` or `MockConsumer`. 

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,23 @@ In case you use podman for testcontainers, some tips must be taken into account:
 
 ---
 
+## Field injection and Parameter Resolution
+
+The extension supports injecting clusters and clients:
+- into fields of the test class 
+- as parameters to `@BeforeAll`
+- as parameters to `@BeforeEach`
+- as parameters to test methods
+
+To avoid collisions with other extensions, such as Mockito, we will only inject into fields which:
+- have no annotations  
+OR are annotated with annotations from the following packages
+- `io.kroxylicious`
+- `org.junit`
+- `java.lang`
+
+The extension will not try to overwrite fields which have already been initialised. 
+
 ## Template tests
 
 You can also use test templates to execute the same test over a number of different cluster configurations. Here's an example:
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
@@ -538,7 +538,9 @@ public class KafkaClusterExtension implements
         boolean supported = true;
         for (Annotation annotation : parameter.getAnnotations()) {
             final String canonicalName = annotation.annotationType().getCanonicalName();
-            if (!canonicalName.startsWith("io.kroxylicious") && !canonicalName.startsWith("org.junit")) {
+            if (!canonicalName.startsWith("io.kroxylicious")
+                    && !canonicalName.startsWith("org.junit")
+                    && !canonicalName.startsWith("java.lang")) {
                 supported = false;
                 break;
             }
@@ -590,9 +592,18 @@ public class KafkaClusterExtension implements
                 .filter(entry -> clientType.isAssignableFrom(entry.getKey()))
                 .map(Map.Entry::getValue)
                 .flatMap(List::stream)
+                .filter(field -> {
+                    try {
+                        return makeAccessible(field).get(testInstance) == null;
+                    }
+                    catch (IllegalAccessException e) {
+                        ExceptionUtils.throwAsUncheckedException(e);
+                    }
+                    return false;
+                })
                 .forEach(field -> {
                     try {
-                        makeAccessible(field).set(testInstance,
+                        field.set(testInstance,
                                 clientFactory.getClient(
                                         "field " + field.getName(),
                                         field,
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
@@ -610,7 +610,7 @@ public class KafkaClusterExtension implements
                 .flatMap(List::stream)
                 .filter(field -> {
                     try {
-                        return field.get(testInstance) == null;
+                        return makeAccessible(field).get(testInstance) == null;
                     }
                     catch (IllegalAccessException e) {
                         ExceptionUtils.throwAsUncheckedException(e);
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
@@ -109,7 +109,7 @@ import static org.junit.platform.commons.util.ReflectionUtils.makeAccessible;
  *         producer.send(new ProducerRecord<>("hello", "world")).get();
  *     }
  * }
- * }**</pre>
+ * }</pre>
  *
  * <p>Notes:</p>
  * <ol>
@@ -121,6 +121,22 @@ import static org.junit.platform.commons.util.ReflectionUtils.makeAccessible;
  * <li>Your test methods can declare {@code Producer}, {@code Consumer} and {@code Admin}-typed parameters.
  * They will be configured to bootstrap against the {@code cluster}.</li>
  * </ol>
+ *
+ * <p>The extension supports injecting clusters and clients:</p>
+ * <ul>
+ *     <li>into fields of the test class</li>
+ *     <li>as parameters to {@code  @BeforeAll}</li>
+ *     <li>as parameters to {@code @BeforeEach}</li>
+ *     <li>as parameters to test methods</li>
+ * </ul>
+ * To avoid collisions with other extensions, such as Mockito, we will only inject into fields which:
+ *  <ul>
+ *      <li> have no annotations</li>
+ *      <li style="list-style: none">OR are annotated with annotations from the following packages</li>
+ *      <li>{@code java.lang}</li>
+ *      <li>{@code org.junit}</li>
+ *      <li>{@code io.kroxylicious}</li>
+ * </ul>
  */
 public class KafkaClusterExtension implements
         ParameterResolver, BeforeEachCallback,
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
@@ -5,10 +5,28 @@
  */
 package io.kroxylicious.testing.kafka.junit5ext;
 
-import io.kroxylicious.testing.kafka.api.KafkaCluster;
-import io.kroxylicious.testing.kafka.api.KafkaClusterConstraint;
-import io.kroxylicious.testing.kafka.api.KafkaClusterProvisioningStrategy;
-import io.kroxylicious.testing.kafka.api.KroxyliciousTestInfo;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Objects;
+import java.util.ServiceLoader;
+import java.util.UUID;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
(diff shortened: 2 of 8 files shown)


SamBarker:
I still need to update the change log

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
```diff
@@ -529,6 +512,39 @@ public void beforeEach(ExtensionContext context) throws Exception {
...
+     * We want to avoid conflicts with annotations such as mockito's @Mock. However, maintaining a list of
+     * conflicting annotations would be mad. So it seems simpler to maintain a set of known safe annotations with which
+     * we can inject still inject.
+     */
+    private static boolean hasOnlySupportedAnnotations(AnnotatedElement parameter) {
+        boolean supported = true;
+        for (Annotation annotation : parameter.getAnnotations()) {
+            final String canonicalName = annotation.annotationType().getCanonicalName();
+            if (!canonicalName.startsWith("io.kroxylicious") && !canonicalName.startsWith("org.junit")) {
+                supported = false;
```
k-wall, line 542:
```suggestion
                supported = false;
                break;
```

k-wall:
> It also prevents injection into fields which already have a non null value.

I don't see the code enforcing this.  Here's a failing test

```java
    @Order(4)
    static Consumer<String, String> mockConsumer = new MockConsumer<>(null);

    @Test
    void fieldWithExistingValueNotInjected() {
        assertTrue(mockConsumer instanceof MockConsumer<String, String>);
    }
```

SamBarker:
Hmm looks like I've failed to push a commit. Thanks for checking @k-wall

k-wall:
I'd add a test ensure that fields with unrecognised annotation don't get injected too.  This test passes.

```
    @Deprecated // poor choice of annotation.
    static Consumer<String, String> fieldWithUnrecognizedAnnotations;

    @Test
    void fieldWithUnrecognizedAnnotationsNotInjected() {
        assertNull(fieldWithUnrecognizedAnnotations);
    }
```

SamBarker:
> I'd add a test ensure that fields with unrecognised annotation don't get injected too. This test passes.
> 
> ```
>     @Deprecated // poor choice of annotation for test case.
>     static Consumer<String, String> fieldWithUnrecognizedAnnotations;
> 
>     @Test
>     void fieldWithUnrecognizedAnnotationsNotInjected() {
>         assertNull(fieldWithUnrecognizedAnnotations);
>     }
> ```

Yeah thats a good point, but I think `java.lang` annotations should probably be allowed. e.g. `@Deprectated` and `@SupressWarnings` should still be valid targets.

junit5-extension/src/test/java/com/example/StubAnnotation.java
```diff
@@ -0,0 +1,11 @@
+package com.example;
```
SamBarker, line 1: Debatable packaging choice, but it needs to be something outside of `io.kroxylicious`

README.md
```diff
@@ -105,6 +105,23 @@ In case you use podman for testcontainers, some tips must be taken into account:
 
 ---
 
+## Field injection and Parameter Resolution
```
k-wall, line 108: Thank you for updating the docs.  Not for action now - wonder if this kind of thing should be package doc?

SamBarker: I can add it in the morning and merge after assuming your happy with a copy and paste of this (give or take formatting)

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/InstanceFieldExtensionTest.java
```diff
@@ -5,27 +5,125 @@
...
+
+    @SuppressWarnings("DeprecatedIsStillUsed")
+    @Deprecated
+    Admin fieldWithJavaLangAnnotations;
+
+    @Name("kafkaCLuster")
+    Admin namedAdmin;
+
+    @Test
+    void shouldInjectConsumerField() {
```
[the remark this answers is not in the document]
k-wall: I think the tests you have added are great improvement.  I would have actually had less confidence if you'd used a parameterized test in this case.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/InstanceFieldExtensionTest.java
```diff
@@ -8,24 +8,94 @@
...
+
+    @Test
+    void shouldInjectIntoFieldWithJunitAnnotation() {
+        assertThat(fieldWithOrderAnnotation).isNotNull().isInstanceOf(Producer.class);
+        // TODO should we use the producer to send a record?
+    }
+
+    @Test
+    void fieldWithUnrecognizedAnnotationsNotInjected() {
+        assertNull(fieldWithUnrecognizedAnnotation);
```
k-wall, line 94: You've. mostly adopted assertThat.. might as well make the remaining tests use the same.
```suggestion
        assertThat(fieldWithUnrecognizedAnnotation).isNull();
```

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/StaticFieldExtensionTest.java
```diff
@@ -48,47 +47,47 @@ public void testKafkaClusterStaticField()
     }
 
     @Test
-    public void adminStaticField() throws ExecutionException, InterruptedException {
+    void adminStaticField() throws ExecutionException, InterruptedException {
```
k-wall, line 50: I don't like to see tests enumerating their checked exceptions. It is just clutter.  In general I prefer to see `throws Exception` when testing an API that throws a checked exception.  However, in this case, I'd have  `assertSameCluster` throw a RuntimeException.

junit5-extension/src/test/java/com/example/StubAnnotation.java
k-wall: I agree that adding the annotation is the right thing to do.

junit5-extension/src/test/java/com/example/StubAnnotation.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface StubAnnotation {
```
k-wall, line 15: Stub makes me think of stubbing code.  How about RuntimeAnnotationUsedForTestPurpose.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/InstanceFieldExtensionTest.java
```diff
@@ -8,24 +8,94 @@
...
+    Consumer<String, String> injectedConsumer;
+
+    Admin injectedAdmin;
+
+    Producer<String, String> injectedProducer;
+
+    @Order(1)
+    Producer<String, String> fieldWithOrderAnnotation;
+
+    final MockConsumer<String, String> mockConsumer = new MockConsumer(OffsetResetStrategy.LATEST);
```
k-wall, line 47:
```suggestion
    final MockConsumer<String, String> mockConsumer = new MockConsumer<>(OffsetResetStrategy.LATEST);
```

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/InstanceFieldExtensionTest.java
```diff
@@ -8,24 +8,94 @@
...
 
 @ExtendWith(KafkaClusterExtension.class)
-public class InstanceFieldExtensionTest extends AbstractExtensionTest {
+class InstanceFieldExtensionTest extends AbstractExtensionTest {
 
     @BrokerCluster(numBrokers = 1)
     KafkaCluster instanceCluster;
 
+    @BrokerCluster(numBrokers = 1)
+    @Name("kafkaCLuster")
```
k-wall, line 35: Funny capitalisation

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
```diff
@@ -555,36 +574,45 @@ private void injectFields(ExtensionContext context, Object testInstance, Class<?
...
+                .map(Map.Entry::getValue)
+                .flatMap(List::stream)
+                .filter(field -> {
                     try {
-                        makeAccessible(field).set(testInstance, getAdmin(
-                                "field " + field.getName(),
-                                field,
-                                field.getType().asSubclass(Admin.class),
-                                context));
+                        return field.get(testInstance) == null;
```
k-wall, line 597: The code now takes an inconsistent approach to visibility.  You need to mutate the visibility before you check for nullness.  Add a test case that checks that private fields are supported too.

```
java.lang.IllegalAccessException: class io.kroxylicious.testing.kafka.junit5ext.KafkaClusterExtension cannot access a member of class io.kroxylicious.testing.kafka.junit5ext.InstanceFieldExtensionTest with modifiers "private"
```

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/StaticFieldExtensionTest.java
```diff
@@ -18,15 +19,13 @@
...
+import java.util.concurrent.ExecutionException;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 
 @ExtendWith(KafkaClusterExtension.class)
-public class StaticFieldExtensionTest extends AbstractExtensionTest {
+class StaticFieldExtensionTest extends AbstractExtensionTest {
 
     @Order(1)
```
k-wall, line 30: I'm not sure what the order annotations are actually trying to achieve.  My first thought was that it was a mechanism to ensure that the field where populated in a particular order, but the code ensures that that clusters are created before clients. 

As far as i can see, the @Order annotation  is actually have no effect.

README.md
k-wall: Second thoughts - maybe extending the javadoc on KafkaClusterExtension would be a better home for this text?

k-wall (approved):
Couple of nits, but happy to approve.  Nice work @SamBarker

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/StaticFieldExtensionTest.java
k-wall: Maybe @tombentley knows the rationale for the `@Order` annotation in the test.

SamBarker: I don't think its about controlling element creation/injection. Its controlling junit evaluation order, so I think its trying to say that if `1` fails there is no hope of anything else passing so evaluate `1` first to give a stable baseline. 

I don't know if it delivers anything in practice however (now we have a functional extension, it was probably quite helpful in bootstrapping everything).

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/StaticFieldExtensionTest.java
SamBarker: Yeah, I think `assertSameCluster` should catch and re-throw as runtime exceptions. However I don't want to pull that thread in this PR. I'm far enough down the rabbit hole from where I started without doing that right now.

junit5-extension/src/test/java/com/example/StubAnnotation.java
SamBarker: I went with `RuntimeMarkerAnnotation` in dba0e5e0db10f70e0b6126117378ca2f9ddf494a as it was a bit shorter and conveyed the important point. 

I think the `ForTestPurposes` bit is implicit in being `com.example` in `src/test`.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/InstanceFieldExtensionTest.java
SamBarker: fixed in 478820f0cee6e7fc36c98987032076ff00c4e258

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
SamBarker: Good point. Addressed in e65ca1ba1d2f14e0a2a6e2ff87f57836bf1f890b.

README.md
SamBarker: Added it to `KafkaClusterExtension` (in 450c030824cca96ceda8acba2cb221dc535de0a2) as that is the natural entry point for users. 

It fits well with the existing notes.

robobario (approved):
LGTM

PR #190, opened 2023-10-03
Allow controller-only nodes in container mode

robobario:
### Type of change

- Enhancement / new feature

### Description
We can now specify `@KraftCluster(numControllers=2)` and ` @BrokerCluster(numBrokers = 1)` in `TEST_CLUSTER_EXECUTION_MODE=CONTAINER` where `numControllers > numBrokers` where previously we could only do this in IN_VM mode.

Why:
https://github.com/ozangunalp/kafka-native/issues/88 was merged and the docker image now supports running in controller only mode. We disable auto configuration if the node has no broker role to prevent kafka-native configuring any roles or listeners, using only the configuration file we prepare.

Note:
Had to move where we determine the roles in KafkaClusterConfig#ConfigHolder because the process.roles are inserted into the map after ConfigHolder construction. Needs cleaning with fire.

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


impl/src/test/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactoryTest.java
```diff
@@ -8,14 +8,13 @@
...
 import static org.junit.jupiter.api.Assertions.fail;
 
 class KafkaClusterFactoryTest {
 
     @SuppressWarnings("resource")
     @Test
-    void shouldThrowInContainerModeWithControllerOnlyNodes() {
+    void shouldCreateInstanceInContainerModeWithControllerOnlyNodes() {
         // Due to https://github.com/ozangunalp/kafka-native/issues/88 we can't support controller only nodes, so we need to fail fast
         // Given
```
k-wall, line 19:
```suggestion
        // Given
```

impl/src/test/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactoryTest.java
```diff
@@ -26,7 +25,12 @@ void shouldThrowInContainerModeWithControllerOnlyNodes() {
                 .build();
 
         // When
-        assertThrows(IllegalStateException.class, () -> KafkaClusterFactory.create(kafkaClusterConfig));
+        try (var kafkaCluster = KafkaClusterFactory.create(kafkaClusterConfig)) {
+            assertThat(kafkaCluster).isNotNull();
+        }
+        catch (Exception e) {
```
k-wall, line 31: the catch is redundant, tests that throw uncaught exceptions fail anyway.  This pattern just needlessly wraps the exception.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -226,6 +217,10 @@ private KafkaContainer buildKafkaContainer(KafkaClusterConfig.ConfigHolder holde
             kafkaContainer.addFixedExposedPort(holder.getExternalPort(), CLIENT_PORT);
             kafkaContainer.addFixedExposedPort(holder.getAnonPort(), ANON_PORT);
         }
+        else {
+            // disables automatic configuration of broker role and default listeners by kafka-native
+            kafkaContainer.withEnv("SERVER_AUTO_CONFIGURE", "false");
```
k-wall, line 222: what's stopping us setting this flag in the `holder.isBroker() == true` case?  what are we missing from the configuration?

k-wall:
> Had to move where we determine the roles in KafkaClusterConfig#ConfigHolder because the process.roles are inserted into the map after ConfigHolder construction. 

yuck, that's violates least surprise.  I haven't absorbed sufficient code to make a suggestion, but I really think this needs to be made sane somehow.

robobario:
> > Had to move where we determine the roles in KafkaClusterConfig#ConfigHolder because the process.roles are inserted into the map after ConfigHolder construction. 
> 
> yuck, that's violates least surprise.  I haven't absorbed sufficient code to make a suggestion, but I really think this needs to be made sane somehow.
> 

It feels like we should be able to model what types of roles we want per node and then derive the properties. 

But getting there is quite twisted. For example a broker needs to know some ports from all controller nodes. So when do you resolve all the ports? How do you gather the ports from each node which has the controller trait?

k-wall:
> > > Had to move where we determine the roles in KafkaClusterConfig#ConfigHolder because the process.roles are inserted into the map after ConfigHolder construction.
> > 
> > 
> > yuck, that's violates least surprise.  I haven't absorbed sufficient code to make a suggestion, but I really think this needs to be made sane somehow.
> 
> It feels like we should be able to model what types of roles we want per node and then derive the properties.
> 
> But getting there is quite twisted. For example a broker needs to know some ports from all controller nodes. So when do you resolve all the ports? How do you gather the ports from each node which has the controller trait? Add in dimensions like "I need to talk to zookeeper instead". So a bit of work to model.

Yes, you are right. It feels like a significant rework.  Maybe doing this isn't so important right now after all.

k-wall (approved):
couple of nits, but lgtm

SamBarker (approved):
LGTM.

Thanks for following through to fix this @robobario.

re: 
> It feels like a significant rework. Maybe doing this isn't so important right now after all.

Yeah its a big task that is definitely separate to this PR. We certainly lack a topology model that could be used to determine what nodes to deploy and what properties each node has. Fixing deserves some thought.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -192,32 +191,26 @@ public TestcontainersKafkaCluster(DockerImageName kafkaImage, DockerImageName zo
...
-                            + clusterConfig.getBrokersNum() + " brokers and " + clusterConfig.getKraftControllers() + " controllers");
-        }
-    }
-
     @NotNull
     private KafkaContainer buildKafkaContainer(KafkaClusterConfig.ConfigHolder holder) {
         String netAlias = "broker-" + holder.getBrokerNum();
+        Properties properties = new Properties();
+        properties.putAll(holder.getProperties());
+        properties.put("log.dir", getBrokerLogDirectory(holder.getBrokerNum()));
```
SamBarker, line 199: Why is this set here a not in the holder? Why is it special in test containers? Shouldn't each container have the same path?

robobario: TestContainers has a mounted in volume `/kafka` so that the log dir can outlive the container getting restarted, so for broker-0 it's logdir is set to `/kafka/broker-0/`.

The InVm cluster creates a temp dir on the local machine and sets `log.dir` to `${tempDir}/broker-0`.

SamBarker: Ok, but shouldn't the log dir still be consistent between them and thus set in the config holder?

robobario: Having a go at refactoring now.

robobario: Have pushed up a refactor that centralizes generation of the per-broker log dir configuration. 

I'm not sure if it's an improvement, there was an interface `KafkaEndpoints` being used to describe the public endpoints, which is an environment specific behaviour. So I tried to generalize this to describe other characteristics of the Cluster when run with a particular execution mode. I called it `MaterializedKafkaCluster`, but this doesn't feel great, so any advances on that would help. 

I also considered adding more interfaces like KafkaEndpoints (something like KafkaStorage) but then there's a lot of signatures that would need both things and it feels like they should be packaged up somehow.

As per the other chain, if this is too deep I'll revert and we can improve it separately.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/InstanceFieldExtensionTest.java
```diff
@@ -5,10 +5,8 @@
  */
 package io.kroxylicious.testing.kafka.junit5ext;
 
-import com.example.RuntimeMarkerAnnotation;
-import io.kroxylicious.testing.kafka.api.KafkaCluster;
-import io.kroxylicious.testing.kafka.common.BrokerCluster;
-import io.kroxylicious.testing.kafka.invm.InVMKafkaCluster;
+import java.util.concurrent.ExecutionException;
```
k-wall, line 8: raised https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/191

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -602,22 +604,26 @@ public ConfigHolder(Properties properties, Integer externalPort, Integer anonPor
...
         }
     }
 
     /**
-     * The interface Kafka endpoints.
+     * Describes aspects of the KafkaCluster that are only material in a particular environment,
+     * for example after port allocation or creation of temporary directories.
      */
-    public interface KafkaEndpoints {
+    public interface MaterializedKafkaCluster {
```
k-wall, line 626: This class doesn't actually need to know about MaterializedKafkaCluster so think a pull-up to the top level is in order. It does need to know about Listener, so it should stand at the top too.  See my PR.

k-wall, line 626: This class doesn't actually need to know about MaterializedKafkaCluster so think a pull-up to the top level is in order. It does need to know about Listener, so it should stand at the top too.  See my PR.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
k-wall: I see you did this. resolving comment.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall, line 626: I'm not keen on the MaterializedKafkaCluster name. Maybe RuntimeClusterConfiguration would be better.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
k-wall: The refactoring definitely represents an improvement.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -46,16 +45,14 @@
 
 import io.kroxylicious.testing.kafka.api.KafkaCluster;
 import io.kroxylicious.testing.kafka.api.TerminationStyle;
-import io.kroxylicious.testing.kafka.common.KafkaClusterConfig;
-import io.kroxylicious.testing.kafka.common.PortAllocator;
-import io.kroxylicious.testing.kafka.common.Utils;
+import io.kroxylicious.testing.kafka.common.*;
```
robobario, line 48: how did this sneak in?

SamBarker: nit: I think we avoid `*` imports. I'm surprised that passes the build.

SamBarker (approved):
I like the initial commit and am happy to see that merged. 

I think my comment on the initial commit about logDirs and the way @robobario has taken the refactoring commit point at real smells. I'm not convinced that `RuntimeClusterConfiguration` is the right thing (it could be). 

I think we should split them into separate PR's and iterate a bit on the refactoring.

robobario:
Cool, I'll move the refactor into another PR and take a look at the broader problems with passing maps/properties everywhere

PR #196, opened 2023-10-09
Refactor modelling of Kafka Topology

robobario:
### Type of change

- Refactoring

### Description

Instead of handing over the user's desired KafkaClusterConfig and letting each driver manifest that into a cluster, we introduce a higher level idea of a KafkaClusterTopology. The idea is we can describe what types of nodes we want (controller/broker/combined) in a topology regardless of what implementation we are going to use to manifest that into some running processes. The topology 

Defining what roles/nodes we want up front means that we don't have to use janky mechanisms to determine what the roles are, like:
1. looking at what ports a process has exposed
2. consulting the properties which we will pass to the kafka server
3. comparing the node id to the quantity requested by the user (plus consider any additional nodes that were added)

Introduces:
1. KafkaDriver: this is the connection to the InVM and Testcontainers implementations that we use to create and destroy kafka nodes and zookeepers. It also supplies endpoint configuration.
5. KafkaTopology: this represents the (mutable) set of kafka nodes (and optionally zookeeper) and tracks their current state. It is dynamic and is used to stop/start/remove nodes. This implements the KafkaCluster interface.
6. KafkaNodeConfiguration: this represents the runtime configuration of a node. Some of the configuration will be pulled from the user specified KafkaClusterConfig and some will be obtained from the driver (endpoints)
7. KafkaNode: this represents the deployed node. This is an abstraction that is implemented by each driver. It can be started/stopped/removed.
8. ZookeeperConfiguration: as above this is the runtime configuration of zookeeper. Currently just it's endpoint details obtained from the driver.
9. Zookeeper: as above, this is an abstraction of the zookeeper process implemented by each each driver.

KafkaNodeConfiguration is responsible for generating the properties for the server using it's knowledge of what roles the node has. The responsibilities of ConfigHolder are now owned by KafkaNodeConfiguration (making certain ports and endpoint addresses accessible). It can consult it's parent topology to discover information about the other nodes (for example to build up the quorum address list).

### Additional Context
We've noticed in several PRs that it feels bad using the nodeId combined with the user's desired quantity of controllers and brokers to try and determine the roles of the node (or inspecting the ports exposed). This along with the tangle of maps being passed around made it tricky to reason about. For instance, recently when enabling controller only mode we had to be careful where we inferred from the generated properties that the node was a controller. This change makes the topology of the cluster a part of the model so that we can designate it's roles and other characteristics up front. We won't need to look at it's properties to determine if it's a broker or controller because it will be an immutable property of that node.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaNode.java
```diff
@@ -0,0 +1,28 @@
...
+import io.kroxylicious.testing.kafka.api.TerminationStyle;
+
+public interface KafkaNode {
+
+    /**
+     * Stops the node and returns how long in milliseconds any associated resources like zookeeper sessions live for
+     * @param terminationStyle termination style
+     * @return milliseconds until orphaned resources expire
+     */
+    int stop(TerminationStyle terminationStyle);
```
robobario, line 17: this was a bit sad, we had a specialization in testcontainers land to sleep for the zookeeper session timeout after stopping some nodes.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaOrchestrator.java
```diff
@@ -0,0 +1,147 @@
...
+                TimeUnit.SECONDS,
+                getNumOfBrokers());
+        return node.nodeId();
+    }
+
+    @Override
+    public void removeBroker(int nodeId) throws UnsupportedOperationException, IllegalArgumentException, IllegalStateException {
+        KafkaNode node = kafkaTopology.get(nodeId);
+        if (node == null) {
+            throw new IllegalArgumentException("nodeId isn't in topology: " + nodeId);
```
SamBarker, line 57: I realise most of this is actually just shuffling code around but...

I'm wondering if we should let this happen and just return. Why should we prevent people calling remove a nodeId that is already gone? 

I can imagine calling `removeBroker` if `addBroker` threw an exception to ensure it wasn't in some zombie state. 

`node` being `null` here represents the desired state.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaTopology.java
```diff
@@ -0,0 +1,237 @@
...
+        kafkaTopology.addNodes(nodes);
+        return kafkaTopology;
+    }
+
+    @NotNull
+    private static Map<Integer, KafkaNodeConfiguration> generateConfigurations(KafkaDriver driver, KafkaClusterConfig config, KafkaTopology kafkaTopology) {
+        int nodesToCreate = Math.max(config.getBrokersNum(), config.getKraftControllers());
+        Stream<KafkaNodeConfiguration> nodeConfigurationStream = IntStream.range(0, nodesToCreate)
+                .mapToObj(value -> new IdAndRoles(roles(config, value), value))
+                // note that generation of a configuration may allocate resources like ports to it
```
SamBarker, line 90: I feel like this comment should be in the javadoc for the method or at call sites such as line 75.

Or is it meant to convey that `getNodeConfiguration` might do the allocation? In which case this maybe the right location for the comment but `getNodeConfiguration` should probably be re-named

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaTopology.java
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    public void startZookeeperIfRequired() {
+        if (zookeeper != null) {
+            zookeeper.start();
+        }
+    }
+
+    public record IdAndRoles(Set<Role> roles, int nodeId) {
```
SamBarker, line 122: Calling it `IdAndRoles` and then having the arguments as `Roles, id` really messes with me...

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaTopology.java
```diff
@@ -0,0 +1,237 @@
...
+        }
+        else if (nodeId < config.getBrokersNum() && nodeId < config.getKraftControllers()) {
+            return Set.of(Role.BROKER, Role.CONTROLLER);
+        }
+        else if (nodeId < config.getBrokersNum()) {
+            return Set.of(Role.BROKER);
+        }
+        else {
+            return Set.of(Role.CONTROLLER);
+        }
```
SamBarker, line 142: nit: I think BROKER should be the default case not Controller.

```suggestion
        else if (nodeId < config.getKraftControllers()) {
            return Set.of(Role.CONTROLLER);
        }
        else {
            return Set.of(Role.BROKER);
        }
```

impl/src/main/java/io/kroxylicious/testing/kafka/common/Role.java
```diff
@@ -0,0 +1,21 @@
...
+    CONTROLLER("controller");
+
+    private final String configString;
+
+    Role(String configString) {
+        this.configString = configString;
+    }
+
+    public String getConfigString() {
+        return configString;
```
SamBarker, line 19: nit

```suggestion
        return name().toLowerCase(Locale.ROOT);
```

Saves passing in the string

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaNodeConfiguration.java
```diff
@@ -0,0 +1,327 @@
...
+     *
+     * @return broker configuration.
+     */
+    @NotNull
+    private Properties generateConfig() {
+        Properties nodeConfiguration = new Properties();
+        nodeConfiguration.putAll(config.getBrokerConfigs());
+        putConfig(nodeConfiguration, "broker.id", nodeIdString());
+
+        var protocolMap = new TreeMap<String, String>();
```
robobario, line 110: These maps of listener properties were also a frustration I didn't get to, still to be cleaned up in a future PR.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaDriver.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public Zookeeper createZookeeper(ZookeeperConfig zookeeperConfig) {
+        return new InVMZookeeper(zookeeperConfig.getPort(), tempDirectory);
+    }
+
+    @Override
+    public void close() {
+        if (tempDirectory.toFile().exists()) {
```
SamBarker, line 71: How does this relate to when the nodes are stopped? 

As its not really safe for this to happen until all nodes are stopped.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaKafkaNode.java
```diff
@@ -0,0 +1,177 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Modifier;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.kafka.common.utils.Time;
+import org.jetbrains.annotations.NotNull;
```
SamBarker, line 19: nit: we should use the findbugs annotations.

 ```suggestion
import edu.umd.cs.findbugs.annotations.NonNull;
```

**Review by SamBarker — COMMENTED** [review]
Thanks @robobario. Shuffling all this around is proving as complicated as I was worried it was...

I'm struggling to understand the relationship and division of responsibilities between a orchestrator and a topology. I will need to take another pass once I grok that distinction.

However as I understand things this is certainly going in the right direction. 

One other question to keep in mind, is how hard would it be to add a Strimzi Driver? What would that look like? (For the record I am in no way suggesting we should implement one in this PR)

**Comment by robobario** [comment]
> I'm struggling to understand the relationship and division of responsibilities between a orchestrator and a topology. I will need to take another pass once I grok that distinction.

Yeah, maybe they should just be combined. I had it in mind that the topology would be lower level, exposing the ability to add/remove/start/stop nodes, but with no safety when it comes to things like stopping brokers before controllers. The orchestrator is supposed to handle these operations that span multiple nodes, stopping things in the right order.

File: impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaKafkaNode.java (hunk 9 above)
Comment (robobario) [reply impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaKafkaNode.java:19]: :( need to set up IDEA properly, I keep getting these inserted automatically on method extract.

File: impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaDriver.java (hunk 8 above)
Comment (robobario) [reply impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaDriver.java:71]: the orchestrator looks after the shutdown:

```
        stopNodesAndDelay(value -> true, TerminationStyle.GRACEFUL, false);
        for (KafkaNode node : kafkaTopology.pureBrokersFirst().toList()) {
            kafkaTopology.remove(node);
        }
        kafkaTopology.close();
        driver.close();
```
so I guess we would javadoc it to say the driver can expect all the nodes to be removed first.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaKafkaNode.java
SamBarker: Here's one I made earlier https://github.com/kroxylicious/kroxylicious/pull/632/commits/c771ce542099019984def3f9ac8fc7b6125e42bd

SamBarker:
> I had it in mind that the topology would be lower level, exposing the ability to add/remove/start/stop nodes, but with no safety when it comes to things like stopping brokers before controllers

I don't think that quite works in terms of a division of responsibilities. When you first mentioned a Topology and an Orchestrator I imagined that there would be things like a ZookeeperTopology or  KRaftTopology.

robobario:
> One other question to keep in mind, is how hard would it be to add a Strimzi Driver? What would that look like? (For the record I am in no way suggesting we should implement one in this PR)

Hmm, one thing that might be an odd fit is the idea of configuring the servers directly. Currently a node configuration can be asked for the properties for a node. The properties are consistent over InVM and testcontainers, just with some runtime differences around the endpoints.

With Strimzi I imagine we wouldn't want that level of control and would just declare how many nodes we need (not sure what the API is looking like for declaring I need X controllers and Y brokers).

So yeah maybe having the topology at the top is kind of counter to the idea of telling Strimzi to just build us a cluster and have it handle some of the details about node id generation etc.

robobario:
> > I had it in mind that the topology would be lower level, exposing the ability to add/remove/start/stop nodes, but with no safety when it comes to things like stopping brokers before controllers
> 
> I don't think that quite works in terms of a division of responsibilities. When you first mentioned a Topology and an Orchestrator I imagined that there would be things like a ZookeeperTopology or KRaftTopology.

I tried to have orchestrator be the one that knows about zookeeper,

SamBarker:
> >I don't think that quite works in terms of a division of responsibilities. When you first mentioned a Topology and an Orchestrator I imagined that there would be things like a ZookeeperTopology or KRaftTopology.

> I tried to have orchestrator be the one that knows about zookeeper,

Yeah I think that makes more sense than my initial thought.

robobario:
Actually, Strimzi should still work. The `KafkaCluster` is still the user facing API and the `StrimziKafkaCluster` implementation would have a very different implementation. I don't think this node orchestration is relevant to Strimzi.

k-wall:
thanks @robobario. I haven't had chance to absorb the details of the PR, but I'm supportive of the refactoring idea.   It has grown into a beast.

SamBarker:
@robobario I'm assuming you've paged this one out, is it worth keeping around or shall we just close the PR?

robobario:
Yep, turned into a bit of a swamp.

PR #199, opened 2023-10-10
Fix broken link in releasing instructions.

k-wall:
### Type of change

_Select the type of your PR_

- Documentation

### Description

Fixing broken link in release instructions.

### Additional Context

_Why are you making this pull request?_

### Checklist

_Please go through this checklist and make sure all applicable tasks have been done_

- [ ] Write tests
- [ ] Make sure all tests pass
- [x] Update documentation
- [ ] Reference relevant issue(s) and close them after merging
- [ ] Update CHANGELOG.md

PR #202, opened 2023-10-11
Improve releasing instructions related to Nexus.

k-wall:
### Type of change

_Select the type of your PR_

- Documentation

### Description

Improve releasing instructions related to Nexus.

Why: this change was prompted by these comments https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/199#discussion_r1353287860 

### Additional Context

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -32,13 +32,13 @@ These are described below.
 
 ### Validating the staged artefacts
 
-You can validate the staged artefacts by having an test application (we'll call it `T`) that uses kroxylicious-junit5-extension use the Maven artefacts by making
+You can validate the staged artefacts by having a test application (we'll call it `T`) that uses kroxylicious-junit5-extension use the Maven artefacts by making
 temporary (local) changes to its POM.
 
 1. Find the staging repository in the [Nexus UI](https://s01.oss.sonatype.org/). It'll be named `iokroxylioustesting-nn`.
 2. [Close](https://help.sonatype.com/repomanager2/staging-releases/managing-staging-repositories) the staging respository.  This gives the staging repository a publicly accessible URL.
 3. Add a [`<repositories>`](https://maven.apache.org/pom.html#Repositories) that references the staging repository public url to `T`'s POM.
-4. Update `T`'s kroxylicious-junit5-extension dependenct to refer to the `<RELEASE_VERSION>`.
+4. Update `T`'s kroxylicious-junit5-extension dependency to refer to the `<RELEASE_VERSION>`.
 5. Run `T` build/test cycle but use an alternative cache location to be sure artefacts are being fetched.  Check the build output, you'll see the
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -21,8 +21,50 @@ the release script to actually perform the release and prepare main for the next
...
+The script will push the Maven artefacts to a Nexus **staging repository** and it will open a PR that will commit the release to Git.
+
+The next steps are:
+1.  Validate the staged artefacts.
+2.  If the staging artefacts are **good**, release the stage repository.  If they are not, **drop** the stage repository and start the release over.
+3.  Merge the PR.
+
+These are described below.
+
+### Validating the staged artefacts
```
k-wall, line 33: We could have a Git Action for this bit.

SamBarker: Yeah that would be very cool, but for another day.

RELEASING.md
```diff
@@ -21,8 +21,50 @@ the release script to actually perform the release and prepare main for the next
...
+
+You can validate the staged artefacts by having a test application (we'll call it `T`) that uses kroxylicious-junit5-extension use the Maven artefacts by making
```
SamBarker, line 35: Should we just have a stub project in GitHub that we use for validating the release?

I had loosely hoped at one stage to use the Sample plug-in for the task but I think I had missed the close step being required for publicly accessible staging repo URL

RELEASING.md
```diff
@@ -21,8 +21,50 @@ the release script to actually perform the release and prepare main for the next
...
+
+### Dropping the stage repository
+
+If the validation step fails and the maven artefacts are deemed bad, drop the staging repository.
+
+Use the [Drop](https://help.sonatype.com/repomanager2/staging-releases/managing-staging-repositories) to drop the stage repository.
+
+In this case the PR should be closed without merging.
+
+Once the issue is resolved, repeat the release, bumping the release micro by 1.
```
SamBarker, line 65: `X.Y.Z` release would not have happened is there any reason to bump Z by 1? We can just fix the issue and re-run the release process can't we?

SamBarker (approved):
Great write up, thanks  @k-wall

RELEASING.md
k-wall: Stub projects (one that consumes the junit ext and one that consumes the filter api) would make sense I think, but I haven't given it too much thought.

RELEASING.md
k-wall: The tag has already been pushed to the origin. The choice is either bump the micro or be prepared to force the tag.  If we go the force tag route, I'd write the script to be behave conservatively and only force the tag if given an extra argument (which would be surfaced at the Git Action level and default to false).

k-wall: I think we can decide later.

Issue #197, opened 2023-10-10
Tight coupling of kafka version


k-wall:
## Describe the bug

There is a tight coupling of the kafka version between the this project and components that use it.  This project has a dependency on several kafka components, which becomes a transitive dependency in the user's projects. 

We saw the situation below where the user's project had a dependency on a single kafka artefact (kafka-client) at a different version.  This lead to a classpath stramash and a runtime failure as an expected symbol was missing.

```
[ERROR] io.kroxylicious.test.tester.KroxyliciousTestersTest.shouldReturnClosableProducer(KafkaCluster) -- Time elapsed: 0.002 s <<< ERROR!
> org.junit.jupiter.api.extension.ParameterResolutionException: Failed to resolve parameter 
```

## Workaround

The user could use dependency management overrides to control all the kafka artefacts to the same version to workaround.  It is a bit opaque as the user wouldn't be entirely certain which artefacts to dependency managed. 

## Proposed Solution

It would probably be better if this project used `provided` dependencies on the kafka artefacts, and insist the user define the kafka version they want to use.  The documentation for this component would need to call out which artefacts are required (kafka-server, -storage etc).
 
              > Upgrading to a version of the junit extension (0.6.0-SNAPSHOT) fixes the parameter resolution error. Lets get the extension released and then revisit the tight coupling of versions.
> 
> ```
> [ERROR] io.kroxylicious.test.tester.KroxyliciousTestersTest.shouldReturnClosableProducer(KafkaCluster) -- Time elapsed: 0.002 s <<< ERROR!
> org.junit.jupiter.api.extension.ParameterResolutionException: Failed to resolve parameter [io.kroxylicious.testing.kafka.api.KafkaCluster cluster] in method [void io.kroxylicious.test.tester.KroxyliciousTestersTest.shouldReturnClosableProducer(io.kroxylicious.testing.kafka.api.KafkaCluster)]: Could not initialize class kafka.server.KafkaConfig$
> 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
> 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
> Caused by: java.lang.NoClassDefFoundError: Could not initialize class kafka.server.KafkaConfig$
> 	at kafka.server.KafkaConfig.LogDirProp(KafkaConfig.scala)
[...]

PR #209, opened 2023-10-24
Fix #143: Default native kafka image tag from the version of Kafka on the classpath

### LINKED ISSUE #143: Default reference to :latest means that user's tests aren't repeatable.
Currently, the framework use a default `:latest` tag in its references to the zookeeper and kafka native images.  As a user of the test framework, I expect test repeatability by default.  The latest tag means this isn't the case. A new version of kafka could be containerized, which could break spontaneously break tests.

The user already has the ability to override the version via the `Version` annotation, but this onerous on the user.   

https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/3cd6816bfa71b0961e6eb7d4616af8d038c4b0be/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java#L74

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Default native kafka image tag from the version of Kafka on the classpath.

Previously TestcontainerKaftaCluster used image tag `latest-snapshot` which meant tests we using the last image build from kafka-natives CI - not at all stable. 

why:
This change, coupled with #197. allows the user of the junit extension a
single way to control the version of the broker used by extension, regardless
of whether they use in-vm, container or both.

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,14 +13,14 @@ Please enumerate all user-facing changes using format `<githib issue/pr number>:
 
 ### Changes, deprecations and removals
 
-* It is now necessary to declare `org.apache.kafka:kafka_2.13` as a test dependency.  If you wish to use kafka in-vm with
-  zookeeper, org.apache.zookeeper:zookeeper must be declared as test dependency too.
+* It is now necessary to declare `org.apache.kafka:kafka_2.13` as a test dependency.
+* If you wish to use kafka in-vm with zookeeper, `org.apache.zookeeper:zookeeper` must be declared as test dependency too.
 * The way `TestcontainersKafkaCluster` determines the native image version tag to use for kafka and zookeeper is changed.
   Previously `TestcontainersKafkaCluster` always used the `latest-snapshot` version.  With this release, it will now default
   to use the version of the kafka native image that corresponds to the version of Kafka Broker found on the classpath.
-  This will give consumers of the test extension that have tests using TestcontainersKafkaCluster and InVMKafkaCluster
-  version consistency and test repeatability.  To get back the original behaviour, annotate the  `KafkaCluster` with
-  `@Version("latest-snapshot")`.
+  This will give consumers of the test extension that have tests using `TestcontainersKafkaCluster` test repeatability
+  and consumers who use both `InVMKafkaCluster` and `TestcontainersKafkaCluster` kafka broker version consistency.
+  To get back the original behaviour, annotate the  `KafkaCluster` with `@Version("latest-snapshot")`.
 
 ## 0.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,20 @@ Please enumerate all user-facing changes using format `<githib issue/pr number>:
 
 ## 0.7.0
 
+* [#143](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/143): Default native kafka image tag from the version of Kafka on the classpath.
 * [#197](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/197): Consumer of the extension should be able to choose the version of the kafka dependency.
+
 
 ### Changes, deprecations and removals
 
-* It is now necessary to declare org.apache.kafka:kafka_2.13 as a test dependency.  If you wish to use kafka in-vm with zookeeper, org.apache.zookeeper:zookeeper
-  must be a test dependency too.
-
+* It is now necessary to declare `org.apache.kafka:kafka_2.13` as a test dependency.  If you wish to use kafka in-vm with
+  zookeeper, org.apache.zookeeper:zookeeper must be declared as test dependency too.
+* The way `TestcontainersKafkaCluster` determines the native image version tag to use for kafka and zookeeper is changed.
+  Previously `TestcontainersKafkaCluster` always used the `latest-snapshot` version.  With this release, it will now default
+  to use the version of the kafka native image that corresponds to the version of Kafka Broker found on the classpath.
+  This will give consumers of the test extension that have tests using TestcontainersKafkaCluster and InVMKafkaCluster
+  version consistency and test repeatability.  To get back the original behaviour, annotate the  `KafkaCluster` with
+  `@Version("latest-snapshot")`.
 
 ## 0.6.0
 
--- a/README.md
+++ b/README.md
@@ -26,14 +26,14 @@ Maven
     <dependency>
         <groupId>org.apache.kafka</groupId>
         <artifactId>kafka_2.13</artifactId>
-        <version>${kafka.version}</version> <!-- versions 3.3.0 or higher is known to work -->
+        <version>${kafka.version}</version> <!-- versions from 3.3.0 have been tested  -->
         <scope>test</scope>
     </dependency>
     <!-- Optional, required if you want to use in-VM kafka in Zookeeper mode. -->
     <dependency>
         <groupId>org.apache.zookeeper</groupId>
         <artifactId>zookeeper</artifactId>
-        <version>${zookeeper.version}</version> <!-- version 3.6.3 or higher is known to work -->
+        <version>${zookeeper.version}</version> <!-- versions from 3.6.3 have been tested -->
         <scope>test</scope>
     </dependency>
 </dependencies>
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -41,6 +41,11 @@
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka_2.13</artifactId>
         </dependency>
+        <!-- Dependency owing to use of org.apache.kafka.server.common.MetadataVersion.MINIMUM_BOOTSTRAP_VERSION -->
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-server-common</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.apache.zookeeper</groupId>
             <artifactId>zookeeper</artifactId>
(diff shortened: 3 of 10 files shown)


k-wall:
This change builds on #197.  Let's review that change first.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -159,10 +160,17 @@ public TestcontainersKafkaCluster(KafkaClusterConfig clusterConfig) {
...
      */
     public TestcontainersKafkaCluster(DockerImageName kafkaImage, DockerImageName zookeeperImage, KafkaClusterConfig clusterConfig) {
-        setDefaultKafkaImage(clusterConfig.getKafkaVersion());
+        var tag = getContainerImageVersionTag(clusterConfig.getKafkaVersion());
+
+        var actualKafkaImage = Optional.ofNullable(kafkaImage).orElse(DEFAULT_KAFKA_IMAGE);
+        var actualZookeeperImage = Optional.ofNullable(zookeeperImage).orElse(DEFAULT_ZOOKEEPER_IMAGE);
+        if (tag != null) {
+            actualKafkaImage = actualKafkaImage.withTag(tag);
+            actualZookeeperImage = actualZookeeperImage.withTag(tag);
```
k-wall, line 169: It's slightly weird that the zookeeper native images are tagged with the kafka version, but that's really a concern for elsewhere.  Zookeeper will wither away anyway 🥀.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -276,13 +277,34 @@ private synchronized String buildBrokerList(Function<Integer, KafkaClusterConfig
...
     /**
      * Gets kafka version.
      *
      * @return the kafka version
      */
     public String getKafkaVersion() {
-        return kafkaImage.getVersionPart();
+        var v = kafkaImage.getVersionPart();
+        if (v != null) {
+            v = v.replaceFirst("^latest-kafka-", "");
```
k-wall, line 305: I've kept the `latest-kafka-` prefix a private concern of the implementation.  Technically this changes the `getKafkaVersion` API, but if you squint at it, you can convince yourself it was a bug all along.

SamBarker (approved):
LGTM

README.md
```diff
@@ -16,11 +16,27 @@ testImplementation 'io.kroxylicious.testing:testing-junit5-extension:0.1'
...
+    <dependency>
+        <groupId>io.kroxylicious.testing</groupId>
+        <artifactId>testing-junit5-extension</artifactId>
+        <version>0.1</version>
+    </dependency>
+    <!-- Kafka Broker version you want to use -->
+    <dependency>
+        <groupId>org.apache.kafka</groupId>
+        <artifactId>kafka_2.13</artifactId>
+        <version>${kafka.version}</version> <!-- versions 3.3.0 or higher is known to work -->
```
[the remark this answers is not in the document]
k-wall: Fixed.

impl/pom.xml
```diff
@@ -41,10 +41,6 @@
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka_2.13</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka-server-common</artifactId>
```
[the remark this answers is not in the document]
k-wall: Yes, you are right

org\.apache\.kafka\.(?!(clients|common))

shows:

import static org.apache.kafka.server.common.MetadataVersion.MINIMUM_BOOTSTRAP_VERSION;

in InVmKafkaCluster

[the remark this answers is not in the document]
k-wall: I've re-added the dependency.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactory.java
```diff
@@ -97,6 +97,10 @@ private static KafkaClusterExecutionMode getExecutionMode(KafkaClusterConfig clu
                 clusterConfig.getExecMode() == null ? KafkaClusterExecutionMode.IN_VM : clusterConfig.getExecMode());
     }
 
+    private static String getKafkaVersion(KafkaClusterConfig clusterConfig) {
+        return System.getenv().getOrDefault(KAFKA_VERSION, clusterConfig.getKafkaVersion() == null ? "latest" : clusterConfig.getKafkaVersion());
```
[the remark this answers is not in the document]
k-wall: I've changed `clusterConfig.getKafkaVersion()` so that it nevers return nulls.
We should now never go down the "latest" path unless someone does an explicit `@Version("latest")`.

k-wall:
I made a snapshot with this PR and tested it with with kroxylicious to make sure that it does indeed pickup the right image if the kafka.broker property is overridden:

We've got no tests in the suite that use TestcontainersKafkaCluster at present, so changing TestcontainersKafkaCluster so that it does and running:

```
mvn  verify -Dkafka.version=3.5.1 -DskipUTs=true -Dit.test="io.kroxylicious.proxy.multitenant.MultiTenantIT#createAndDeleteTopic"  -Dfailsafe.failIfNoSpecifiedTests=false
```

You see that the 3.5.1 image is indeed used.

franvila (approved):

k-wall:
@SamBarker would you be able to re-review?

CHANGELOG.md
```diff
@@ -7,13 +7,20 @@ Please enumerate all user-facing changes using format `<githib issue/pr number>:
...
+* [#143](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/143): Default native kafka image tag from the version of Kafka on the classpath.
 * [#197](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/197): Consumer of the extension should be able to choose the version of the kafka dependency.
 
-### Changes, deprecations and removals
 
-* It is now necessary to declare org.apache.kafka:kafka_2.13 as a test dependency.  If you wish to use kafka in-vm with zookeeper, org.apache.zookeeper:zookeeper
-  must be a test dependency too.
+### Changes, deprecations and removals
 
+* It is now necessary to declare `org.apache.kafka:kafka_2.13` as a test dependency.  If you wish to use kafka in-vm with
```
SamBarker, line 16: It could just be my poor comprehension skills but I think this should really be two paragraphs.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -76,9 +77,13 @@ public class KafkaClusterConfig {
 
     /**
      * Kafka version to be used for deploying kafka in container mode, e.g. "3.3.1".
+     * Defaults to the version available on the classpath.
+     * <br/>
+     * This value is ignored if execMode is not CONTAINER.
```
SamBarker, line 82: I think the double negative is a confusing, personally I'd say something to the effect 

`The value is only used when execMode is CONTAINER`

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -571,6 +576,11 @@ public String clusterId() {
         return isKraftMode() ? kafkaKraftClusterId : null;
     }
 
+    private static String detectKafkaVersionFromClasspath() {
+        var version = AppInfoParser.getVersion();
+        return version == null || version.equals("unknown" /* AppInfoParser.DEFAULT_VALUE */) ? Version.LATEST_RELEASE : version;
```
SamBarker, line 581: Did you mean to leave the reference to AppInfoParser?

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -276,13 +280,45 @@ private synchronized String buildBrokerList(Function<Integer, KafkaClusterConfig
...
     }
 
+    /**
+     * Gets the container image version tag for the specified kafka version.
+     * This knows about the naming conventions used by the ogunalp native image project.
+     *
+     * @param kafkaVersion desired kafka version, which may be null.
+     * @return image tag or null
+     */
+    @Nullable
```
SamBarker, line 290: I'm a bit confused as to what a null return value is supposed to mean. Particularly in the latest release case.

SamBarker (approved):
Mostly nits.

Thanks @k-wall

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: Are you referring to the comment `/* AppInfoParser.DEFAULT_VALUE */`?  My preference would have been to use their constant, AppInfoParser.DEFAULT_VALUE, but that's package private.  I stuck in the comment with the intent of making it clean where "unknown" comes from.

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
SamBarker: Yeah I don't think we need to offer an i dependant override for zookeeper, though I do fully understand your line of thinking.

k-wall: I was thinking about the use-case where a developer wants to override say the kafka native with her own, but leave the zookeeper one using the default.   But like I say, I really don't think this is for this PR or now.

PR #213, opened 2023-10-29
Release script automation

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Automates the existing release process.

The idea is to have two workflows that the developer would run consecutively.

1. stage the release to nexus
2. deploys the stage release to Maven Central.

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


.github/workflows/stage_release.yml
```diff
@@ -0,0 +1,109 @@
...
+    steps:
+      - name: 'Check out repository'
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch }}
+          token: ${{ secrets.KROXYLICIOUS_RELEASE_TOKEN }}
+
+      - name: 'Configure Git username/email'
+        run: |
+          git config user.name kroxylicious
```
k-wall, line 69: Fix me.

.github/workflows/stage_release.yml
```diff
@@ -0,0 +1,109 @@
...
+          ./scripts/release.sh -k ${SIGNING_KEY_SHORT_NAME} \
+                               -v ${{ github.event.inputs.release-version }} \
+                               -b ${{ github.event.inputs.branch }} \
+                              ${{ github.event.inputs.dry-run && '-d' || '' }} \
+                              ${{ github.event.inputs.skip-tests && '-s' || '' }}
+        env:
+          MAVEN_USERNAME: ${{ vars.KROXYLICIOUS_SONAR_TYPE_TOKEN_USERNAME }}
+          MAVEN_CENTRAL_TOKEN: ${{ secrets.KROXYLICIOUS_SONAR_TYPE_TOKEN_PASSWORD }}
+          MAVEN_GPG_PASSPHRASE: ${{ secrets.KROXYLICIOUS_RELEASE_GPG_PASSPHRASE }}
+
```
k-wall, line 101: Should drop repo on failure.

.github/workflows/stage_release.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: 'Close repo'
+        if: ${{ github.event.inputs.dry-run == 'false' }}
+        run: |
+          # Maven deploy leaves the Nexus Staging Repo open, let's close it
+          PLUGIN=org.sonatype.plugins:nexus-staging-maven-plugin:1.6.13
+          # Parse the repo_id from the table rendered by rc-list
+          STAGING_REPO_ID=$(mvn ${PLUGIN}:rc-list -DnexusUrl=https://s01.oss.sonatype.org/ -DserverId=ossrh | awk 'f{print $2;f=0} /.INFO.*ID.*State.*Description/{f=1}')
```
k-wall, line 108: fail reasonable if we don't have a repo id.

.github/workflows/stage_release.yml
k-wall: Done

.github/workflows/stage_release.yml
k-wall: Done.

pom.xml
```diff
@@ -604,6 +604,10 @@
                                 <configuration>
                                     <!--suppress UnresolvedMavenProperty -->
                                     <keyname>${releaseSigningKey}</keyname>
+                                    <gpgArguments>
```
k-wall, line 607: Without this, gpg won't read the passphrase from the environment.   Even with this, interactive use works fine (at least on a Mac).

.github/workflows/stage_release.yml
```diff
@@ -0,0 +1,126 @@
...
+          SONAR_TYPE_TOKEN_USERNAME: ${{ vars.KROXYLICIOUS_SONAR_TYPE_TOKEN_USERNAME }}
+          SONAR_TYPE_TOKEN_PASSWORD: ${{ secrets.KROXYLICIOUS_SONAR_TYPE_TOKEN_PASSWORD }}
+        run: ./scripts/transition-staging-repository-state.sh -s drop
+
+      - name: 'Close stage repository'
+        if: ${{ github.event.inputs.dry-run == 'false' }}
+        env:
+          SONAR_TYPE_TOKEN_USERNAME: ${{ vars.KROXYLICIOUS_SONAR_TYPE_TOKEN_USERNAME }}
+          SONAR_TYPE_TOKEN_PASSWORD: ${{ secrets.KROXYLICIOUS_SONAR_TYPE_TOKEN_PASSWORD }}
+        run: ./scripts/transition-staging-repository-state.sh -s close
```
k-wall, line 125: So, we could omit this `close` step if we switched the pom.xml to use `org.sonatype.plugins:nexus-staging-maven-plugin` rather than the neutral `maven-deploy-plugin`.    We'd then configure nexus-staging-maven-plugin to [skipStagingRepositoryClose=false](https://github.com/sonatype/nexus-maven-plugins/blob/main/staging/maven-plugin/README.md#:~:text=staging%20repository%20(boolean)-,skipStagingRepositoryClose,-false).   I think we'd still want the verification step, so we'd still need the transition-staging-repository-state.sh to handle the other state transitions.

The disadvantage of using nexus-staging-maven-plugin in the POM is that we then tie the project to Nexus.

.github/workflows/deploy_release.yml
```diff
@@ -0,0 +1,77 @@
...
+# This workflow is intended to be run after successfully validation of the release
+# staged by stage_release.  It will release (or drop) the staging repository.
+#
+#
+# It requires the following:
+# variables:
+# KROXYLICIOUS_SONAR_TYPE_TOKEN_USERNAME - Sonar Type Access User Token Username
+# secrets:
+# KROXYLICIOUS_SONAR_TYPE_TOKEN_PASSWORD - Sonar Type Access User Token Password
+# KROXYLICIOUS_RELEASE_TOKEN             - GitHub PAT wih content/createPullRequest permission for Kroxylicious repo.
```
SamBarker, line 28: Should it also have permissions to create a release?

.github/workflows/deploy_release.yml
```diff
@@ -0,0 +1,77 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      next-state:
+        description: 'Desired next state for the staging repository'
+        required: true
+        type: choice
+        default: drop
+        options:
+          - drop
```
SamBarker, line 40: is `close` missing from here for a particular reason?

scripts/transition-staging-repository-state.sh
```diff
@@ -0,0 +1,74 @@
...
+  esac
+done
+
+# The rc-list table looks like this:
+#[INFO] ID                   State    Description
+#[INFO] iokroxylicious-1032  CLOSED   unknown
+#[INFO] ------------------------------------------------------------------------
+
+STAGING_REPO_IDS=$(mvn ${PLUGIN}:rc-list "${MVN_ARGS[@]}" 2>/dev/null \
+                 | awk '/\[INFO\] ---*/{found=0} found==1{print $2} /\[INFO\] ID.*State.*Description/{found=1}')
```
SamBarker, line 45: Yuck. thats a horrible thing to need.

.github/workflows/deploy_release.yml
```diff
@@ -0,0 +1,77 @@
...
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+
+      - name: 'Set up Java'
+        uses: actions/setup-java@v3
+        with:
+          java-version: 17
+          distribution: temurin
+          server-id: ossrh
+          server-username: SONAR_TYPE_TOKEN_USERNAME # env variable for Sonar Type username
```
SamBarker, line 67: Maven central is hosted by `Sonatype` not `Sonar Type`. Sona**r** is a different business entirely. 

I was confused why we needed credentials for SonarCloud here for minute.

.github/workflows/stage_release.yml
```diff
@@ -0,0 +1,126 @@
...
+#
+# Once the staged artefacts have been verified, run the workflow deploy_release with the argument `release` to
+# release the artefacts to Maven Central and merge the release PR.
+#
+# It requires the following:
+# variables:
+# KROXYLICIOUS_SONAR_TYPE_TOKEN_USERNAME - Sonar Type Access User Token Username
+# secrets:
+# KROXYLICIOUS_RELEASE_TOKEN             - GitHub PAT wih content/createPullRequest permission for Kroxylicious repo.
+# KROXYLICIOUS_RELEASE_GPG_PRIVATE_KEY   - GPG private key for Kroxylicious
```
SamBarker, line 31: `GPG private key for Kroxylicious` sounds like it contradicts the description of which GPG Key to use from the PR description. 

I presume this should really be the Private key of the project admin conducting the release.

.github/workflows/stage_release.yml
```diff
@@ -0,0 +1,126 @@
...
+# KROXYLICIOUS_RELEASE_GPG_PASSPHRASE    - GPG passphrase
+# KROXYLICIOUS_SONAR_TYPE_TOKEN_PASSWORD - Sonar Type Access User Token Password
```
SamBarker, line 33: Again, for the record, the company is `Sonatype` not `Sonar Type`

.github/workflows/stage_release.yml
SamBarker: We are tied to Nexus as that is what runs Maven central, that said I think we should stick to the deploy plugin.

RELEASING.md
```diff
@@ -2,69 +2,85 @@
...
 
-Firstly ensure that you have completed the [release requirements](https://github.com/kroxylicious/kroxylicious/blob/main/RELEASING.md#requirements).  Once that is done, use
-the release script to actually perform the release and prepare main for the next development version.
 
-```shell
-./scripts/release.sh -k <YOUR_KEY_SHORT_ID> -v <RELEASE_VERSION>
-```
+| Secret                               | Description                                   |
+|--------------------------------------|-----------------------------------------------|
+| KROXYLICIOUS_RELEASE_GPG_PRIVATE_KEY | GPG private key for Kroxylicious (plain text) |
```
SamBarker, line 24: I think we should be explicit here that its the key of the *person* running the release and not the project.
```suggestion
| KROXYLICIOUS_RELEASE_GPG_PRIVATE_KEY | GPG private key for signing Kroxylicious (plain text) |
```

RELEASING.md
```diff
@@ -2,69 +2,85 @@
...
 
 1. Find the staging repository in the [Nexus UI](https://s01.oss.sonatype.org/). It'll be named `iokroxylioustesting-nn`.
-2. [Close](https://help.sonatype.com/repomanager2/staging-releases/managing-staging-repositories) the staging respository.  This gives the staging repository a publicly accessible URL.
-3. Add a [`<repositories>`](https://maven.apache.org/pom.html#Repositories) that references the staging repository public url to `T`'s POM.
-4. Update `T`'s kroxylicious-junit5-extension dependency to refer to the `<RELEASE_VERSION>`.
-5. Run `T` build/test cycle but use an alternative cache location to be sure artefacts are being fetched.  Check the build output, you'll see the
-   kroxylicious-junit5-extension come from the staging location. 
+2. Add a [`<repositories>`](https://maven.apache.org/pom.html#Repositories) that references the staging repository public url to `T`'s POM.
+3. Update `T`'s kroxylicious-junit5-extension dependency to refer to the `<RELEASE_VERSION>`.
+4. Run `T` build/test cycle but use an alternative cache location to be sure artefacts are being fetched.  Check the build output, you'll see the
```
SamBarker, line 53: Probably worth adding a sample command line for that.

SamBarker (approved):
A very thorough job @k-wall good work.

.github/workflows/deploy_release.yml
k-wall: Possibly, it'd be really easy to have deploy_release call jreleaser to create the release notes and upload them to Git Hub. At that point it would need the permissions.   I think doing this a separate PR would be best.

.github/workflows/deploy_release.yml
k-wall: `stage_release.yml` is closing the release itself, so the developer would never have manually perform that state transition.

scripts/transition-staging-repository-state.sh
k-wall: Indeed, I couldn't believe the maven plugin doesn't give a programmatic way to retrieve the information.   I think they are assuming you stage the release using their plugin which gives you control of the id, so you never need to discover it.

.github/workflows/deploy_release.yml
k-wall: I had the two business conflated in my mind.

RELEASING.md
k-wall: I did add a command line, is this not enough?

```
MAVEN_OPTS="-Dmaven.repo.local=/tmp/repository" mvn verify
```

k-wall:
I gave it another spin against my fork with the changes.
It seems to be working.

Screenshot 2023-11-09 at 17 40 02
Screenshot 2023-11-09 at 17 41 32

RELEASING.md
SamBarker: No that's perfect, it was me missing it in the diff

Issue #219, opened 2023-11-01
Injected KafkaConsumer isn't useable with subscribe gesture.


k-wall:
**Describe the bug**

Injected KafkaConsumer aren't useable with subscribe gesture.  If you try you get an error message like this:

`To use the group management or offset commit APIs, you must provide a valid group.id in the consumer configuration.`

**To Reproduce**
1. Annotate a test method with a KafkaConsumer
2. Include a call to consumer.subscribe(topics);
3. Test fails with exception `To use the group management or offset commit APIs, you must provide a valid group.id in the consumer configuration.`

**Expected behavior**

Ought to be possible to use subscribe.

**Logs**

Attach or copy and paste relevant logs.

**Additional context**

I think we ought to provide a mechanism to set producer/consumer config from an annotation.  That would allow the user to specify a `group.id`.  This would satisfy the requirements of `#subscribe` and be convenient for  other use-cases too.

You can workaround the issue by using the `#assign` gesture instead.

PR #220, opened 2023-11-02
feat: #219 allow kafka configuration to be provided for Producers, Consumer and Admin clients too

k-wall:
### Type of change

_Select the type of your PR_

- Bugfix
- Enhancement / new feature
- Refactoring
- Documentation

### Description

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


franvila (approved):

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaConfig.java
```diff
@@ -0,0 +1,61 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.kroxylicious.testing.kafka.api.KafkaClusterConstraint;
+
+/**
+ * {@code @KafkaConfig} can be used to annotate a field in a test class or a
```
k-wall, line 17: I settled on the name `@KafkaConfig`.      The Kafka prefix is in sympathy with `KafkaCluster` type that is already part of the API. `@Config` would have been too general.

`KafkaConfig` does collude with `kafka.server.KafkaConfig` but given this is annotation I don't think it'll cause too much pain as the user probably won't be importing it anyway.   Why's that?  The constants defined by `kafka.server.KafkaConfig` aren't actually usable within the annotation declaration  This is because they are Scala constants which manifest in Java, as methods.

You'd like to write
```java
@KafkaConfig(name = kafka.server.KafkaConfig.DeleteTopicEnableProp(), value = "false")
```
but it won't compile.

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
```diff
@@ -970,13 +971,28 @@ private static Admin getAdmin(String description,
...
                             sourceElement);
-                    return new Closeable<>(sourceElement, cluster.getClusterId(), new KafkaConsumer<>(cluster.getKafkaClientConfiguration(),
+                    return new Closeable<>(sourceElement, cluster.getClusterId(), new KafkaConsumer<>(buildConfig(sourceElement, cluster),
                             keySerializer, valueSerializer));
                 },
                         (Class<Closeable<KafkaConsumer<?, ?>>>) (Class) Closeable.class)
                 .get();
     }
 
+    private static Map<String, Object> buildConfig(AnnotatedElement sourceElement, KafkaCluster cluster) {
```
SamBarker, line 981: Feels a bit odd to need this here and in `impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java`

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
```diff
@@ -970,13 +971,28 @@ private static Admin getAdmin(String description,
...
+        var clientConfig = cluster.getKafkaClientConfiguration();
+        for (Annotation annotation : sourceElement.getAnnotations()) {
+            if (annotation instanceof KafkaConfig.List configList) {
```
SamBarker, line 984: Debatably we should support `@BrokerConfig` here too. Though its of limited value and duration, thoughts?

SamBarker (approved):
I'm happy with the change as it stands but I'm only +0.5 to merging.

I follow and am largely persuaded about the rational for moving from `BrokerConfig` to `KafkaConfig` but I do wonder if we are letting the technical tail wag the dog. KafkaConfig works because all config is key value primitive pairs. So client config and broker config are thus fungible.

However they are conceptually different things. I'm wondering if we should instead be keeping `BrokerConfig` and introducting `KafkaClientConfig`. That would allow us more flexibility in say validating config parameters (I'm not sure we should) because we could tell which type of usage Key Value pair was for.

k-wall:
> I'm happy with the change as it stands but I'm only +0.5 to merging.
> 
> I follow and am largely persuaded about the rational for moving from `BrokerConfig` to `KafkaConfig` but I do wonder if we are letting the technical tail wag the dog. KafkaConfig works because all config is key value primitive pairs. So client config and broker config are thus fungible.
> 
> However they are conceptually different things. I'm wondering if we should instead be keeping `BrokerConfig` and introducting `KafkaClientConfig`. That would allow us more flexibility in say validating config parameters (I'm not sure we should) because we could tell which type of usage Key Value pair was for.

I was torn between the two whilst writing the PR.    If you prefer separation, then that's enough to tip the scales.  I'll go that way.

Would ClientConfig as a name be acceptable (to follow the convention established by BrokerConfig)?

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
k-wall: Moot now, but my intent was to leave the deprecated BrokerConfig as a mechanism for supplying broker config - i.e. leave its semantics unchanged.

SamBarker:
>I was torn between the two whilst writing the PR. If you prefer separation, then that's enough to tip the scales. I'll go that way.

I'm torn as well, but we are both leaning the same way so if your happy lets move to a client config annotation.

> Would ClientConfig as a name be acceptable (to follow the convention established by BrokerConfig)?

`ClientConfig` works for me

k-wall:
@SamBarker Comments addressed in the last commit.

SamBarker (approved):
LGTM

Thanks @k-wall

PR #229, opened 2023-11-13
Issue 221 - ability to inject topics into tests

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Ability to inject test topics into tests. 

The user signals their wish for a topic to be injected by the use of a `Topic` type parameter.  On encountering this, the Junit extension will create a topic on the broker using a random name, and let the test know the name via the `Topic` parameter.    The test is guaranteed that the topic will exist before the test method commences.

Passing topic config, replication factor and number of partitions is supported.

### Additional Context

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md


junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicPartitions.java
```diff
@@ -0,0 +1,28 @@
...
+import io.kroxylicious.testing.kafka.api.KafkaClusterConstraint;
+
+/**
+ * Used to specify the number of partitions of a {@link Topic} injected into
+ * a test as field or parameter.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.PARAMETER })
+@KafkaClusterConstraint
+public @interface TopicPartitions {
```
SamBarker, line 22: I don't have strong views on this but I'd like to hear your thoughts on why partition and replica counts are separate annotations rather than properties of the topic annotation. 

The later feels like a slightly more natural expression to me (probably using nullable fields for inheriting broker defaults)

impl/src/main/java/io/kroxylicious/testing/kafka/internal/AnonymousAdminSource.java
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright Kroxylicious Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.kroxylicious.testing.kafka.internal;
+
+import org.apache.kafka.clients.admin.Admin;
+
+public interface AnonymousAdminSource {
```
SamBarker, line 11: We can always refactor later but why is "Anonymous" part of the API? I see that as, an important, implementation detail. 

To expand on that we could supply a an admin client with fixed credentials which we have also injected into the cluster and satisfy the contract we need from this interface.

SamBarker (approved):
I'm happy in principal with this change. 

I do have a few questions about the exact shape, but they could be equally well addressed in follow up PRs as being fixed before this one merges.

franvila (approved):

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicPartitions.java
k-wall: > I don't have strong views on this but I'd like to hear your thoughts on why partition and replica counts are separate annotations rather than properties of the topic annotation.

My motivation was the cardinality - TopicConfig naturally needs to be repeatable in order to support > 1.

SamBarker: A repeatable topicConfig makes sense but i meant  partitions and replication factor specifically

impl/src/main/java/io/kroxylicious/testing/kafka/internal/AnonymousAdminSource.java
k-wall: Good point, it is an implementation detail.

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
```diff
@@ -1124,11 +1186,11 @@ private void assertSupportedType(String target, Class<?> type) {
     }
 
     @FunctionalInterface
-    private interface ClientFactory<T, X extends T> {
-        X getClient(String description,
-                    AnnotatedElement sourceElement,
-                    Class<X> type,
-                    Type genericType,
-                    ExtensionContext extensionContext);
+    private interface Injector<T, X extends T> {
```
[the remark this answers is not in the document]
k-wall: Done

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicPartitions.java
k-wall: Ah, so you mean introduce a `@Topic` annotation which would carry the replicationFactor/replicationFactor fields.  I presume that actually adding the annotation would remain optional.

So it would look like:

```
@Topic(replicationFactor = 2) Topic myTopic 
```

this form would work too.

```
Topic myTopic 
```

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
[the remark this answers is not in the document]
k-wall: Done. I made it package private.

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicPartitions.java
k-wall: Alternatively,  we could chose to special case the config names `replication.factor` and `partitions`, and use those to pass the configuration.  Those names aren't understood by the Kafka Client, but they do have currency already in Kafka Connect (`org.apache.kafka.connect.runtime.TopicCreationConfig#REPLICATION_FACTOR_CONFIG`).  That would avoid the need for the additional annotation but would mean we have the extra javadoc.

k-wall:
@SamBarker @franvila  I want to point out the package/module location of the new @TopicConfig annotation.  See #234 for more details.

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/Topic.java
```diff
@@ -0,0 +1,18 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * Signals the wish for a topic to be created and injected into a test as either test parameter
+ * or field.
+ * {@see KafkaClusterExtension}
+ */
+public interface Topic {
+    @NonNull
+    String name();
```
k-wall, line 17: I did think about exposing the topicUuid here.  I decided, YAGNI.

SamBarker: Yeah, I think YAGNI is the right call.

It will be very cheap to add if we come up with a motivation for exposing both so lets leave it out for now.

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicPartitions.java
SamBarker: > `@Topic(replicationFactor = 2) Topic myTopic `

Was exactly what I had in mind. I confess `@Topic Topic` does look a little odd. 

I think your right about `@TopicParitions` vs using `@TopicConfig` and handling repetition so the dedicated annotations make more sense than re-using `TopicConfig`.

*My personal* preference would still be an optional `@Topic` but I'm happy to merge this and come revisit later.

k-wall: I have a marginal preference for `@TopicParitions/`@TopicReplicationFactor` over `@Topic Topic` so I'm going to stick with my way.

k-wall:
> @SamBarker @franvila I want to point out the package/module location of the new @TopicConfig annotation. See #234 for more details.

PR is good to merge although could be affected by the outcome of #234.  Holding off merge until that reaches conclusion.

Issue #266, opened 2024-01-12
Release process generates incorrect snapshot version for main when releasing for a previous version from main.


k-wall:
The issue is the same as https://github.com/kroxylicious/kroxylicious/issues/869.

PR #284, opened 2024-03-04
Fix #254: automate release note production

### LINKED ISSUE #254: Release workflow should automate release note production
**Is your feature request related to a problem? Please describe.**

Release workflow should automate release note production, like the kroxylicious repo already does, in order to remove the manual steps.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

There was some accumulated technical debt on the release pipeline. This PR put the mechanism back in step with https://github.com/kroxylicious/kroxylicious.

#254 and #266 

### Additional Context

### Checklist

- [ ] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/scripts/stage-release.sh
+++ b/scripts/stage-release.sh
@@ -6,15 +6,19 @@
 #
 
 set -e
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+. "${SCRIPT_DIR}/common.sh"
 
 REPOSITORY="origin"
 BRANCH_FROM="main"
-SNAPSHOT_INCREMENT_INDEX=2
 DRY_RUN="false"
 SKIP_VALIDATION="false"
 TEMPORARY_RELEASE_BRANCH=""
 PREPARE_DEVELOPMENT_BRANCH=""
 ORIGINAL_GH_DEFAULT_REPO=""
+GPG_KEY=""
+RELEASE_VERSION=""
+DEVELOPMENT_VERSION=""
 while getopts ":v:b:k:r:n:dsh" opt; do
   case $opt in
     v) RELEASE_VERSION="${OPTARG}"
@@ -24,12 +28,8 @@ while getopts ":v:b:k:r:n:dsh" opt; do
     r) REPOSITORY="${OPTARG}"
     ;;
     k) GPG_KEY="${OPTARG}"
-      if [[ -z "${GPG_KEY}" ]]; then
-          echo "GPG_KEY not set unable to sign the release. Please specify -k <YOUR_GPG_KEY>" 1>&2
-          exit 1
-      fi
     ;;
-    n) SNAPSHOT_INCREMENT_INDEX="${OPTARG}"
+    n) DEVELOPMENT_VERSION="${OPTARG}"
     ;;
     d) DRY_RUN="true"
     ;;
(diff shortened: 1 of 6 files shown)


k-wall:
I ran stage then deploy (with the drop argument) against my fork.  All seemed well.

Screenshot 2024-03-04 at 14 38 36
Screenshot 2024-03-04 at 14 39 09
Screenshot 2024-03-04 at 14 39 46
Screenshot 2024-03-04 at 14 41 09
Screenshot 2024-03-04 at 14 42 59

scripts/stage-release.sh
```diff
@@ -102,6 +105,13 @@ cleanup() {
     fi
 }
 
+setVersion() {
+  local VERSION=$1
+  mvn -q -B versions:set -DnewVersion="${VERSION}" -DgenerateBackupPoms=false -DprocessAllModules=true
+
+  git add '**/pom.xml' 'pom.xml'
+}
```
SamBarker, line 113: I've just changed this in the main/runtime repo to 

```suggestion
updateVersions() {
  local FROM_VERSION=$1
  local NEW_VERSION=$2
  mvn -q -B versions:set -DnewVersion="${NEW_VERSION}" -DgenerateBackupPoms=false -DprocessAllModules=true

  # Bump version ref in files not controlled by Maven
  # shellcheck disable=SC2046
  ${SED} -i -e "s#${FROM_VERSION//./\\.}#${NEW_VERSION}#g" $(find kubernetes-examples -name "*.yaml" -type f)

  git add '**/*.yaml' '**/pom.xml' 'pom.xml'
}
```

I know we don't need the `sed` command here but I wonder if we should keep API consistency. 

I'm starting to wonder if we should be baking all this scripting into a github action that both projects use?

scripts/stage-release.sh
```diff
@@ -126,11 +136,14 @@ if [[ "${SKIP_VALIDATION:-false}" != true ]]; then
 fi
 
 echo "Versioning Kroxylicious-junit-extension as ${RELEASE_VERSION}"
+setVersion "${RELEASE_VERSION}"
 
-mvn -q versions:set -DnewVersion="${RELEASE_VERSION}" -DgenerateBackupPoms=false
-echo "Validating things still build"
-mvn -q clean install -Pquick
+#Set the release version in the Changelog
+${SED} -i -e "s_##\sSNAPSHOT_## ${RELEASE_VERSION//./\\.}_g" CHANGELOG.md
```
SamBarker, line 142: This PR misses adding the corresponding `SNAPSHOT` entry to the changelog.

SamBarker (approved):
Thanks @k-wall

scripts/stage-release.sh
k-wall: >I'm starting to wonder if we should be baking all this scripting into a github action that both projects use?

+1  I'd been thinking similar things.

I'd been wondering about moving the release machinery to a separate repo and then reusing it from both projects.

I was thinking to keep the mainstay as scripts that you can run locally.  I say this because AFAIK there's no way to run a workflow from a local machine, so that makes testing script slow (involves GitHub). 

But I'm open to alternative ideas.

SamBarker: It was an off the cuff thought but I wasn't really think of anything more sophisticated than using the action as the distribution mechanism for the scripts (and standardising the args)

scripts/stage-release.sh
k-wall: I verified that the CHANGELOG is updated to include `SNAPSHOT`.

```bash
uthor: kwall <kwall@apache.org>
Date:   Thu Mar 14 11:23:48 2024 +0000

    Start next development version

    Signed-off-by: kwall <kwall@apache.org>

diff --git a/CHANGELOG.md b/CHANGELOG.md
index 9001975..604770b 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Please enumerate all user-facing changes using format `<githib issue/pr number>:

 ## 1.0.0

+## SNAPSHOT
```

Issue #13, opened 2022-12-21
Consider a general @User annotation


tombentley:
(This is some reasoning for why #8 might be a good idea). 

Currently `@User` is explicitly associated with `@SaslPlain`. That's a pattern that we _could_ continue (e.g. @SaslScram might have its own user, and `@SaslOauthBearer` too). But that might be unrealistically restrictive. i.e. we're currently assuming that a provider can actually provision a user _with a particular password_. Some KaaS might allow to provision a user using some SASL mechanism, but _not_ allow you to pick your own password. 

So we should consider separating the _declaration of a user_ from the _credentials required_ to authenticate as that user. The author of a test typically doesn't care what the credentials actually _are_ anyway. What they care about is having a number of distinct accounts (and possibly clients authenticated using those accounts). Having a bare `@User` annotation would also allow us to re-use it on client declarations, so specify that the client should be configured to authenticate as that user, with the provider responsible for generating passwords and supplying the JAAS config fragment needed for configuring the clients. i.e. That would allow us to express a test configuration like this:

```java
@Test myTest(@User("alice") @User("bob") KafkaCluster, // a cluster with users alice and bob
             @User("alice") Producer<String, String> aliceProducer, // a client authenticated as alice
             @User("bob") Consumer<String, String> bobProducer) // a client authenticated as bob
```

The value of this is somewhat lessened unless we're also able to specify the _authorization_ for users. And that gets verbose to do using annotations. i.e. to do this properly we end up needing to have declarative topics and ACLs, something like:

```java
@Test myTest(
             @User("alice") 
             @User("bob") 
             @Topic("myTopic") 
             @TopicAcl(topic="myTopic", principal="alice", operation=WRITE) 
             @TopicAcl(topic="myTopic", principal="bob", operation=READ) 
             KafkaCluster, // a cluster with users alice and bob
             @User("alice") Producer<String, String> aliceProducer, // a client authenticated as alice
             @User("bob") Consumer<String, String> bobProducer) // a client authenticated as bob
```

OTOH if we _didn't_ have something like this, then we end up having to support methods on `KafkaCluster` to let people create users and ACLs in order to support tests like this. 

[...]

PR #330, opened 2024-06-16
Extend annotations to allow SASL mechanisms other than PLAIN

k-wall:
### Type of change

- Enhancement / new feature

### Description

Deprecates `@SaslPlainAuth` annotation replacing it with `@SaslMechanism` annotation.  For mechanisms that need it, a list of principals can be passed.  

```java
@SaslMechanism(value = "PLAIN", principals = { @Principal(user = "alice", password = "foo) } )
```

This takes a steps towards #13 but it does not allow user annotations to be attached to kafka clients.

### Additional Context

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md


junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
```diff
@@ -39,59 +38,52 @@
...
             throws Exception {
         assertSameCluster(clusterWithConfigs, admin);
-        ConfigResource resource = new ConfigResource(ConfigResource.Type.BROKER, "0");
+        var resource = new ConfigResource(ConfigResource.Type.BROKER, "0");
+
         var configs = admin.describeConfigs(List.of(resource)).all().get().get(resource);
-        assertEquals("zstd", configs.get("compression.type").value());
-        assertEquals("false", configs.get("delete.topic.enable").value());
+        assertThat(configs.get("compression.type")).extracting(ConfigEntry::value).isEqualTo("zstd");
+        assertThat(configs.get("delete.topic.enable")).extracting(ConfigEntry::value).isEqualTo("false");
```
SamBarker, line 79: Don't get me wrong I think the switch to assertJ is a good one, but that line still makes me a little sad. It would be good to have a custom assertion or something that understood `configs` is effectively a map...

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
```diff
@@ -126,167 +118,177 @@ public void producerConfiguration(KafkaCluster cluster,
...
+                                            @BrokerCluster(numBrokers = 2) @Name("B") KafkaCluster clusterB,
+                                            @Name("B") Admin adminB,
+                                            @Name("A") Admin adminA)
+            throws Exception {
         assertSameCluster(clusterA, adminA);
-        assertEquals(1, describeCluster(clusterA.getKafkaClientConfiguration()).nodes().get().size());
-        assertEquals(1, describeCluster(adminA).nodes().get().size());
+        assertClusterIdAndSize(clusterA, 1);
+
         assertSameCluster(clusterB, adminB);
```
SamBarker, line 159: Ohh should we be also have `assertSameCluster(clusterA, adminA);`?

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
```diff
@@ -126,167 +118,177 @@ public void producerConfiguration(KafkaCluster cluster,
...
-        var dc = describeCluster(cluster.getKafkaClientConfiguration());
-        assertEquals(1, dc.nodes().get().size());
+    void kraftBasedClusterParameter(@BrokerCluster @KRaftCluster KafkaCluster cluster)
+            throws Exception {
+        assertClusterIdAndSize(cluster, 1);
+    }
+
+    @Test
+    void saslPlainAuth(@BrokerCluster @User(user = "alice", password = "foo") KafkaCluster cluster) {
+        doAuthExpectSucceeds(cluster, "alice", "foo");
```
SamBarker, line 204: hmmm. the repetition of `alice` and `foo` gives me pause. Should we be injecting a User/principal into the test? instead of purely annotating the cluster?

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
```diff
@@ -126,167 +118,177 @@ public void producerConfiguration(KafkaCluster cluster,
...
-            throws ExecutionException, InterruptedException {
-        var dc = describeCluster(cluster.getKafkaClientConfiguration());
-        assertEquals(1, dc.nodes().get().size());
+    void kraftBasedClusterParameter(@BrokerCluster @KRaftCluster KafkaCluster cluster)
+            throws Exception {
+        assertClusterIdAndSize(cluster, 1);
+    }
+
+    @Test
+    void saslPlainAuth(@BrokerCluster @User(user = "alice", password = "foo") KafkaCluster cluster) {
```
SamBarker, line 203:
```suggestion
    void shouldDefaultToSaslPlain(@BrokerCluster @User(user = "alice", password = "foo") KafkaCluster cluster) {
```

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
```diff
@@ -126,167 +118,177 @@ public void producerConfiguration(KafkaCluster cluster,
...
+    void saslPlainAuthDeprecatedAnnotation(@BrokerCluster @User(user = "alice", password = "foo") KafkaCluster cluster) {
+        doAuthExpectSucceeds(cluster, "alice", "foo");
+    }
+
+    @Test
+    @SuppressWarnings({ "deprecated", "java:S4144" })
+    void saslPlainAuthDeprecatedAnnotationManyUsers(@BrokerCluster @SaslPlainAuth(user = "alice", password = "foo") @SaslPlainAuth(user = "bob", password = "bar") KafkaCluster cluster) {
+        doAuthExpectSucceeds(cluster, "alice", "foo");
+        doAuthExpectSucceeds(cluster, "bob", "bar");
+    }
```
SamBarker, line 235: Should we also cover the `List` version of `SaslPlainAuth`?

SamBarker (commented):
I think this is an improvement and mergeable but wanted to take a cycle to discuss the merits or otherwise of scoping users to auth mechanisms.

impl/src/main/java/io/kroxylicious/testing/kafka/common/BrokerCluster.java
```diff
@@ -60,10 +65,9 @@
  *     public void testKafkaClusterParameter(
  *             @BrokerCluster(numBrokers = 3)
  *             @KRaftCluster(numControllers = 3)
- *             @SaslPlainAuth({
- *                 @UserPassword(user="alice", password="foo"),
- *                 @UserPassword(user="bob", password="bar")
- *             })
+ *             @User(user="alice", password="foo")
+ *             @User(user="bob", password="bar")
+ *             @SaslMechanism("SCRAM-SHA-256")
```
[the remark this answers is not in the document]
k-wall: I agree the current APIs are probably a bit off.   I think clusters should probably have `@Listener` and those should have a optional SASL mechanism property and a list of users.    Listeners should probably have a @Name which can then be referred to by an annotation on the injected the Kafka Client.  

```java
@Test
void myTest(@Listener(name="foo", saslMechanism="SCRAM-SHA-512 @User("alice", "secret)) KafkaCluster cluster, @ListenerRef(name="foo", userRef="alice") admin) {
...
}
```

Unfortunately, I don't think we have the time to spend right now on this.  I think what we have is serving us well enough.  Having the ability to test other SASL mechanisms will be a enabler for the Kroxylicious project.

I do hope we get time to revisit this again.  I'd love to see the Junit5 ext get traction outside Kroxylicious.  I think it has got real value for anyone integration testing with Kafka.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -167,61 +177,125 @@ public static KafkaClusterConfig fromConstraints(List<Annotation> annotations, T
...
+
+    private static Optional<Map<String, String>> calculateActualUsers(Optional<Map<String, String>> saslUsers, Optional<Map<String, String>> deprecatedSaslUsers) {
+        if (saslUsers.isPresent()) {
+            if (deprecatedSaslUsers.isPresent()) {
+                throw new IllegalArgumentException("Cannot use deprecated SaslPlainAuth with SaslUser.");
+            }
+            return saslUsers;
+        }
+        else if (deprecatedSaslUsers.isPresent()) {
+            if (DEPRECATED_SASL_PLAIN_AUTH_USE_REPORTED.compareAndExchange(false, true)) {
```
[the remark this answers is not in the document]
k-wall: yes, my intent is to log just once if the use of the deprecated API.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -167,61 +177,125 @@ public static KafkaClusterConfig fromConstraints(List<Annotation> annotations, T
...
+        var deprecatedSaslUsers = Optional.<Map<String, String>> empty();
+
         for (Annotation annotation : annotations) {
             if (annotation instanceof BrokerCluster brokerCluster) {
                 builder.brokersNum(brokerCluster.numBrokers());
             }
-            if (annotation instanceof KRaftCluster kRaftCluster) {
+            else if (annotation instanceof KRaftCluster kRaftCluster) {
                 builder.kraftMode(true);
                 builder.kraftControllers(kRaftCluster.numControllers());
```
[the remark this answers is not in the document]
k-wall: I tried that, I ended up with a lombok/javadoc related to the builder parameter.  I gave up trying to solve that.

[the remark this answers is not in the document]
SamBarker: Ok fair enough!

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
k-wall: `assertSameCluster(clusterA, adminA);` is already part of this test (twoDefinedClusterParameterAndAdmin)

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
k-wall: missing test added.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
k-wall: maybe, i think we can leave that discussion for another day, when there is more time to consider a deeper refactoring or even a v2.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/ParameterExtensionTest.java
k-wall: Fixed.

impl/src/main/java/io/kroxylicious/testing/kafka/common/BrokerCluster.java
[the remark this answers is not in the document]
k-wall: actually the idea of making `@User` a child of `@Listener` is duff.  SCRAM users are scoped globally to the Cluster.  They are not exclusive to the listener.   (For PLAIN, you can have a `bob` with a different password at each listener.  But that's just weird.. it is still the same bob..  maybe we don't need to model that).

making User a child of SaslMechanism would be a better.

@SaslMechanism(mechanism="PLAIN", @User("alice", "secret")...)

that would also leave the door open for a `@Listener` annotation   `@Listener(mechanism="PLAIN" name="foo")`.  I'm not suggesting adding a @Listener now.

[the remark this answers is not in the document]
k-wall: > making User a child of SaslMechanism

@SamBarker what would you think to this idea?

k-wall:
@SamBarker Sonar issues are now addressed.

impl/src/main/java/io/kroxylicious/testing/kafka/common/BrokerCluster.java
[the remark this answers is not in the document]
SamBarker: I like the idea 

```java
@BrokerCluster
@SaslMechanism(mechanism = "SCRAM-SHA-512",
               principals = { @Principal(identity = "alice", secret = "foo"), @Principal(identity = "bob", secret = "bar") })
KafkaCluster cluster
```

Feels like a good (if verbose) model. `SaslMechanism` possibly also needs a name so that we could attach them to listeners if we went down that route (but we should probably apply YAGNI and add it later...)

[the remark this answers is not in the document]
k-wall: Thanks, I'll give that a go

k-wall:
@SamBarker I've refactored along the lines suggested by the comment.  Can you re-review please?

README.md
```diff
@@ -84,7 +84,10 @@ You can configure different clusters by annotating the `KafkaCluster` field or p
 * `@BrokerCluster(numBrokers=3)` will use a Kafka cluster with the given number of brokers
 * `@KRaftCluster` will ensure a KRaft-based cluster is used. `@KRaftCluster(numControllers=3)` will use a controller quorum with 3 controllers.
 * `@ZooKeeperCluster` will ensure a ZooKeeper-based Kafka cluster (unsurprisingly this is mutually exclusive with `@KRaftCluster`)
-* `@SaslPlainAuth` will provide cluster with `SASL-PLAIN` authentication.
+* `@SaslMechanism` will provide cluster with the external listener configured for the given SASL
+   mechanism.  Use of this option requires the client to use SASL authentication. For PLAIN and SCRAM mechanism a
+   database of principals must be provided.
+* `@SaslPlainAuth` will provide cluster with `SASL-PLAIN` authentication (deprecated - use @User).
```
SamBarker, line 90:
```suggestion
* `@SaslPlainAuth` will provide cluster with `SASL-PLAIN` authentication (deprecated - use @SaslMechanism).
```

README.md
```diff
@@ -106,6 +109,20 @@ class MyTest {
...
 
+## Configuring SASL
+
+To config the Broker to use SASL, use the annotations `@SaslMechanism` to specify the SASL
+mechanism. Use `@User` configure a user and password.  The `@User` annotation  may be repeated
+to define multiple users.
+
+```java
+class MyTest {
+@SaslMechanism("PLAIN") @User(user = "alice", password = "foo") KafkaCluster cluster;
```
SamBarker, line 120: this has drifted with the refactor, need to update it to use `principals =`

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -163,65 +172,112 @@ public static boolean supportsConstraint(Class<? extends Annotation> annotation)
...
                 builder.kafkaVersion(version.value());
             }
-            if (annotation instanceof BrokerConfig.List brokerConfigList) {
-                for (var config : brokerConfigList.value()) {
-                    builder.brokerConfig(config.name(), config.value());
-                }
+            else if (annotation instanceof SaslMechanism mechanism) {
+                useSasl = true;
+                builder.saslMechanism(Optional.ofNullable(mechanism.value()).orElse(PLAIN_SASL_MECHANISM_NAME));
+                var principals = Optional.ofNullable(mechanism.principals()).orElse(new SaslMechanism.Principal[]{});
```
SamBarker, line 209: Should this be `orElseThrow` as there seems little point in setting up a cluster with no users? Actually answer my own question there will be a programmatic mechanism to add users to the cluster. 

Maybe we should log a warning if there are no configured principals so thats its clear to users that they need to do something else.

impl/src/test/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfigTest.java
```diff
@@ -125,6 +128,228 @@ void kafkaVersionDisallowsNulls() {
         assertThatThrownBy(() -> kafkaClusterConfigBuilder.kafkaVersion(null)).isInstanceOf(NullPointerException.class);
     }
 
+    @Test
```
SamBarker, line 131: ❤️ Loving the given when then usage!

SamBarker (approved):
Need to update the documentation before merging but the change LGTM.

Thanks for the refactoring @k-wall I think its a big improvment.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: Yes, I wondered about this.   You are correct, in the SCRAM case the user can can users later and that fine.  For the PLAIN, that's not the case.    The API could enforce the latter, or like you say, log an exception.  However this would be a job for the underlying KafkaCluster API rather than the annotation facade.  I'll do this separately.

README.md
k-wall: The duplication (and inconsistencies) between the javadoc and README is irksome.   We should probably just have one or the other.

PR #337, opened 2024-06-19
Create SCRAM users in advance of cluster startup

k-wall:
### Type of change

_Select the type of your PR_

- Bugfix

### Description

Why: avoid race condition that lead to sporadic authentication failures during tests involving SCRAM users.

During development of #330 I was noticing occasional sporadic test fails in KRaft mode with SCRAM users.  The client
would fail with a authentication error.  I reason that the creating SCRAM users via the Admin client does not give the
guarantee that the user is immediately available for authentication (indeed, introducing non-viable  a Thread.sleep() worked around the issue).

Prepopulating the SCRAM users before startup seems like the best way to eliminate the issue.

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/KraftLogDirUtil.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/KraftLogDirUtil.java
@@ -26,6 +26,7 @@ import scala.collection.immutable.Seq;
  */
 final class KraftLogDirUtil {
     private static final PrintStream LOGGING_PRINT_STREAM = LoggingPrintStream.loggingPrintStream(InVMKafkaCluster.LOGGER, System.Logger.Level.DEBUG);
+    private static final String FORMAT_METHOD_NAME = "formatCommand";
     private static final boolean IGNORE_FORMATTED = true;
 
     private KraftLogDirUtil() {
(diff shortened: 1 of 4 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/KraftLogDirUtil.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private static void formatReflectively(Object metaProperties, Seq<String> directories, BootstrapMetadata bootstrapMetadata, MetadataVersion metadataVersion)
+            throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+        var formatMethod = StorageTool.class.getDeclaredMethod("formatCommand", PrintStream.class, Seq.class, metaProperties.getClass(), BootstrapMetadata.class,
+                MetadataVersion.class,
+                boolean.class);
+        // note ignoreFormatter=true so tolerate a log directory which is already formatted. this is
+        // required to support start/stop.
+        formatMethod.invoke(null, LOGGING_PRINT_STREAM, directories, metaProperties, bootstrapMetadata, metadataVersion, true);
```
SamBarker, line 60: nit: extract `ignoreFormatter=true` as a constant. e.g `IGNORE_FORMATTER` or maybe `IGNORE_FORMATTED` so the call is a bit more self describing.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/KraftLogDirUtil.java
```diff
@@ -0,0 +1,89 @@
...
+        catch (Exception e) {
+            throw new RuntimeException("failed to prepare log dirs for KRaft", e);
+        }
+    }
+
+    private static void formatReflectively(Object metaProperties, Seq<String> directories, BootstrapMetadata bootstrapMetadata, MetadataVersion metadataVersion)
+            throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+        var formatMethod = StorageTool.class.getDeclaredMethod("formatCommand", PrintStream.class, Seq.class, metaProperties.getClass(), BootstrapMetadata.class,
+                MetadataVersion.class,
+                boolean.class);
```
SamBarker, line 57: I'm feeling confused as to why this solves the problem of the Storage tool changing packages. We still have the import from (what I presume is the new package). Don't we need to use reflection to load `StorageTool.class` and from that invoke/find the `formatCommand` method? 

Note I think this is an existing issue and not really new in this refactor...

SamBarker (approved):
LGTM 

one tiny nit and something outside of scope for this PR

impl/src/main/java/io/kroxylicious/testing/kafka/invm/KraftLogDirUtil.java
k-wall: It was the return type of buildMetadataProperties that changed.  It was previously kafka.server.MetaProperties but got changed to org.apache.kafka.metadata.properties.MetaProperties

k-wall:
Tested manually against older kafka versions (>= 3.3) to ensure that existing use-cases are maintained.

PR #353, opened 2024-07-08
Enable zk with out kafka cluster

SamBarker:
### Type of change

- Bugfix

### Description

Use reflection hackery to force `enableEntityConfigControllerCheck` to false. 

### Additional Context

We can't configure SCRAM users before starting Kafka without this.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -141,7 +141,7 @@ public class KafkaClusterConfig {
     @Singular
     private final Map<String, String> brokerConfigs;
 
-    @SuppressWarnings("java:S5738") // silence warnings about the use of deprecated code
+    @SuppressWarnings({ "java:S5738", "removal" }) // silence warnings about the use of deprecated code
     private static final Set<Class<? extends Annotation>> SUPPORTED_CONSTRAINTS = Set.of(
             ClusterId.class,
             BrokerCluster.class,
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
@@ -10,6 +10,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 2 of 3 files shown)


showuon (approved):
LGTM! I agree this is the best workaround, compared with using `KafkaZkClient.apply` with some other hack with the private `zkTlsClientAuthEnabled` method, or invoking ConfigCommand to add scram users. I think this did what we want to test, and ZK is going to get removed, we don't have to spend too much time on it. Thanks for the fix!

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -141,7 +141,7 @@ public class KafkaClusterConfig {
     @Singular
     private final Map<String, String> brokerConfigs;
 
-    @SuppressWarnings("java:S5738") // silence warnings about the use of deprecated code
+    @SuppressWarnings({ "java:S5738", "removal" }) // silence warnings about the use of deprecated code
```
k-wall, line 144: I didn't know about "removal"

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -491,4 +493,22 @@ private void createScramUsersInZookeeper(List<UserScramCredentialRecord> parsedC
...
+            final Class<? extends KafkaZkClient> zkClientClass = zkClient.getClass();
+            final Field enableEntityConfigControllerCheck = zkClientClass.getDeclaredField("enableEntityConfigControllerCheck");
+            enableEntityConfigControllerCheck.setAccessible(true);
+            enableEntityConfigControllerCheck.setBoolean(zkClient, false);
+        }
+        catch (NoSuchFieldException ignored) {
+            //presumably we are on kafka <= 3.7.0, so we can move on with life
+        }
+        catch (IllegalArgumentException | IllegalAccessException | SecurityException e) {
+            System.out.println("tried to fix it, went boom");
```
k-wall, line 509: we should use a logger.

k-wall:
lgtm. In another place,  https://github.com/ozangunalp/kafka-native/pull/195/files this was the approach that was taken.  But I'm happy with this approach.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
SamBarker: I'm guessing its intellij specific but still useful to shutup another warning.

PR #355, opened 2024-07-13
Kafka 3.8.0 compatibility

k-wall:
### Type of change

- Enhancement / new feature

### Description

Changes to achieve Kafka 3.8.0 compatibility made in manner to retain compatibility with earlier versions.
(Tested against 3.8.0 RC1)

### Additional Context

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -44,7 +44,6 @@ import org.junit.jupiter.api.TestInfo;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
-import kafka.server.KafkaConfig;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.Singular;
@@ -506,7 +505,7 @@ public class KafkaClusterConfig {
         putConfig(server, "zookeeper.connect", kafkaEndpoints.getEndpointPair(Listener.CONTROLLER, 0).connectAddress());
         putConfig(server, "zookeeper.sasl.enabled", "false");
         putConfig(server, "zookeeper.connection.timeout.ms", Long.toString(60000));
-        putConfig(server, KafkaConfig.ZkSessionTimeoutMsProp(), Long.toString(6000));
+        putConfig(server, "zookeeper.session.timeout.ms", Long.toString(6000));
     }
 
     private void configureKraftNode(KafkaEndpoints kafkaEndpoints, int nodeId, Properties nodeConfiguration, TreeMap<String, String> protocolMap,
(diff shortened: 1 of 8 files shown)


k-wall:
I think it would be useful to merge/release this soon to help those who are testing their component at the 3.8.0 RCs.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -168,7 +168,7 @@ private KafkaConfig buildBrokerConfig(KafkaClusterConfig.ConfigHolder c) {
         Properties properties = new Properties();
         properties.putAll(c.getProperties());
         var logsDir = getBrokerLogDir(c.getBrokerNum());
-        properties.setProperty(KafkaConfig.LogDirProp(), logsDir.toAbsolutePath().toString());
+        properties.setProperty("log.dir", logsDir.toAbsolutePath().toString());
```
SamBarker, line 171: again feels odd to be moving towards hardcoded strings from lookups.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -516,7 +515,7 @@ private void configureKraftNode(KafkaEndpoints kafkaEndpoints, int nodeId, Prope
         putConfig(nodeConfiguration, "node.id", Integer.toString(nodeId)); // Required by Kafka 3.3 onwards.
 
         var quorumVoters = IntStream.range(0, kraftControllers)
-                .mapToObj(controllerId -> String.format("%d@//%s", controllerId, kafkaEndpoints.getEndpointPair(Listener.CONTROLLER, controllerId).connectAddress()))
+                .mapToObj(controllerId -> String.format("%d@%s", controllerId, kafkaEndpoints.getEndpointPair(Listener.CONTROLLER, controllerId).connectAddress()))
```
SamBarker, line 518: Was the format wrong before or did it change as part of 3.8.0? I guess what I'm really getting at is does this need to checked per version somehow.

SamBarker (commented):
@k-wall  I don't really understand why this helps compatibility? Moving from property lookups to hard coded names feels like moving in the wrong direction. 

If it gives us compatibility across Kafka versions then I'm +1 to merging I just don't follow as things stand.

k-wall:
The property methods have been removed in Kafka 3.8.0. Replacements exist in other classes, but I couldn't see a reasonable why to utilise them and retain compile time compatibility.

I want to merge this before 3.8.0 is release so that users (Strimzi) have a release that works against new and old.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: The format used by the Ext for this property was wrong (according to the Kafka documentation).  The validation used by 3.8.0 rejects the slashes.

SamBarker (approved):
Given the lack of compile time property lookups this LGTM

PR #412, opened 2024-11-04
Remove unused site plugin declaration

robobario:
### Type of change

- Refactoring

### Description

While checking a dependabot bump for this plugin I didn't see any output for it in the CI build logs. I think we aren't using it anywhere.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -367,11 +367,6 @@
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-site-plugin</artifactId>
-                    <version>3.20.0</version>
-                </plugin>
-                <plugin>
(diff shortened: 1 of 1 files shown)


robobario:
yarg :

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.5.0:enforce (enforce-plugin-versions) on project testing-parent: 
Error:  Rule 1: org.apache.maven.enforcer.rules.RequirePluginVersions failed with message:
Error:  Some plugins are missing valid versions or depend on Maven 3.8.8 defaults (LATEST, RELEASE as plugin version are not allowed)
Error:     org.apache.maven.plugins:maven-site-plugin. 	The version currently in use is 3.3 via default lifecycle bindings
Error:  -> [Help 1]
```

I guess we want the version under control because it's a default plugin and we're enforcing it, despite it never being invoked?

PR #413, opened 2024-11-04
Remove unused site plugin declaration

robobario:
### Type of change

- Refactoring

### Description

Remove unused plugin so we do not have to maintain it via dependabot

Note that it seems to be detected as an active plugin by the enforcer despite apparently contributing no invocations to a maven install or deploy, so we exclude it from the enforcement.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -367,11 +367,6 @@
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-site-plugin</artifactId>
-                    <version>3.20.0</version>
-                </plugin>
-                <plugin>
-                    <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>3.5.2</version>
                     <configuration>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -498,7 +493,8 @@
                                             <banLatest>true</banLatest>
                                             <banRelease>true</banRelease>
                                             <banSnapshots>false</banSnapshots>
-                                            <phases>clean,deploy,site</phases>
```
SamBarker, line 501: I have no idea why we ever included the `site` phase.

pom.xml
```diff
@@ -498,7 +493,8 @@
...
+                                            <phases>clean,deploy</phases>
+                                            <unCheckedPluginList>org.apache.maven.plugins:maven-site-plugin</unCheckedPluginList>
```
SamBarker, line 497: Do we actually need this after the `site` phase is removed?

SamBarker (approved):

SamBarker:
Do we need something similar in the runtime repo?

pom.xml
robobario: yeah it detects the maven-site-plugin even if phases is set to just `clean` or just `deploy`.

robobario: maybe just because site is so deeply ingrained in maven https://maven.apache.org/ref/3.6.3/maven-core/lifecycles.html#site_Lifecycle it's one of the 3 predefined lifecycles

robobario:
yep, have created the equivalent over in the kroxy repo https://github.com/kroxylicious/kroxylicious/pull/1569

PR #419, opened 2024-11-15
Refactor endpoint creation and allow client connections to be formed to controllers in container mode.

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature
- Refactoring

### Description

Follow up to https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/417.     Allows client connections to be formed to controllers in container mode too.

The problem I faced was that `EndpointPair` was not sufficient to represent of the  requirements of the controller listener in the Container case.

1.  The address that the _client_ needs to connect to is an endpoint exposed by Docker on the host.
2. Other brokers on the cluster need to connect to controller via an address on the Docker network.

This lead me to this refactoring:

 * `KafkaEndpoints` renamed to `KafkaListenerSource` to better reflect role
 * `EndpointPair` renamed to `KafkaListener` and made responsible for bind, kafka network and advertised addresses.

I also slimmed down ConfigHolder and improved the tests.

### Additional Context

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -476,8 +476,8 @@ public class KafkaClusterConfig {
     private static void configureInternalListener(TreeMap<String, String> protocolMap, TreeMap<String, String> listeners, KafkaListener interBrokerEndpoint,
                                                   TreeMap<String, String> advertisedListeners, TreeSet<String> earlyStart, Properties server) {
         protocolMap.put(INTERNAL_LISTENER_NAME, SecurityProtocol.PLAINTEXT.name());
-        listeners.put(INTERNAL_LISTENER_NAME, interBrokerEndpoint.bind().toString());
-        advertisedListeners.put(INTERNAL_LISTENER_NAME, interBrokerEndpoint.advertised().toString());
+        listeners.put(INTERNAL_LISTENER_NAME, interBrokerEndpoint.bind().address());
+        advertisedListeners.put(INTERNAL_LISTENER_NAME, interBrokerEndpoint.advertised().address());
         earlyStart.add(INTERNAL_LISTENER_NAME);
         putConfig(server, "inter.broker.listener.name", INTERNAL_LISTENER_NAME);
     }
@@ -485,21 +485,21 @@ public class KafkaClusterConfig {
     private static void configureAnonListener(TreeMap<String, String> protocolMap, TreeMap<String, String> listeners, KafkaListener anonEndpoint,
                                               TreeMap<String, String> advertisedListeners) {
         protocolMap.put(ANON_LISTENER_NAME, SecurityProtocol.PLAINTEXT.name());
-        listeners.put(ANON_LISTENER_NAME, anonEndpoint.bind().toString());
-        advertisedListeners.put(ANON_LISTENER_NAME, anonEndpoint.advertised().toString());
+        listeners.put(ANON_LISTENER_NAME, anonEndpoint.bind().address());
+        advertisedListeners.put(ANON_LISTENER_NAME, anonEndpoint.advertised().address());
     }
 
     private static void configureExternalListener(TreeMap<String, String> protocolMap, String externalListenerTransport, TreeMap<String, String> listeners,
                                                   KafkaListener clientEndpoint, TreeMap<String, String> advertisedListeners) {
         protocolMap.put(EXTERNAL_LISTENER_NAME, externalListenerTransport);
-        listeners.put(EXTERNAL_LISTENER_NAME, clientEndpoint.bind().toString());
-        advertisedListeners.put(EXTERNAL_LISTENER_NAME, clientEndpoint.advertised().toString());
+        listeners.put(EXTERNAL_LISTENER_NAME, clientEndpoint.bind().address());
+        advertisedListeners.put(EXTERNAL_LISTENER_NAME, clientEndpoint.advertised().address());
     }
 
     private static void configureLegacyNode(KafkaListenerSource kafkaListenerSource, ConfigHolder configHolder) {
         KafkaListener kafkaListener = kafkaListenerSource.getKafkaListener(Listener.CONTROLLER, 0);
         var server = configHolder.properties();
-        putConfig(server, "zookeeper.connect", kafkaListener.kafkaNet().toString());
+        putConfig(server, "zookeeper.connect", kafkaListener.kafkaNet().address());
         putConfig(server, "zookeeper.sasl.enabled", "false");
         putConfig(server, "zookeeper.connection.timeout.ms", Long.toString(60000));
         putConfig(server, "zookeeper.session.timeout.ms", Long.toString(6000));
@@ -516,7 +516,7 @@ public class KafkaClusterConfig {
 
         var quorumVoters = IntStream.range(0, kraftControllers)
                 .mapToObj(controllerId -> String.format("%d@%s", controllerId,
-                        kafkaListenerSource.getKafkaListener(Listener.CONTROLLER, controllerId).kafkaNet()))
+                        kafkaListenerSource.getKafkaListener(Listener.CONTROLLER, controllerId).kafkaNet().address()))
                 .collect(Collectors.joining(","));
         putConfig(nodeConfiguration, "controller.quorum.voters", quorumVoters);
         putConfig(nodeConfiguration, "controller.listener.names", CONTROLLER_LISTENER_NAME);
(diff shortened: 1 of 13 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/ScramUtils.java
```diff
@@ -27,25 +36,272 @@ static ScramCredential asScramCredential(UserScramCredentialRecord uscr) {
...
-        return CollectionConverters.asJava(StorageTool.getUserScramCredentialRecords(n).get());
     }
 
     @NonNull
-    private static String toKafkaScramCredentialsFormat(String username, String password) {
+    private static String toKafkaScramArguments(String username, String password) {
         return "[name=%s,password=%s]".formatted(username, password);
     }
+
+    /** Copied from Kafka 3.9 ScramParser */
```
k-wall, line 70: Verbatim copy of  ScramParser from Kafka 3.9.   The kafka class's accessibility mean we can't use it directly.

k-wall:
I can't easily improve the test coverage.  I make a couple of "chasing coverage" changes to address existing coverage gap by way of compensation. 

 There are a couple of things that could be done - making sure that the tests run by Maven with `TEST_CLUSTER_EXECUTION_MODE=CONTAINER` contribute coverage would be the biggest win coverage wise, but that would be outside the scope of this PR.    

I intend to commit once that review comments are addressed ignoring the Quality Gate.

k-wall:
One thing I didn't think about is how users of the annotations should gester that they want a controller admin connection rather than a regular one.  I think we can leave that until we have a use-case where we need such a thing.

k-wall:
@SamBarker could you add this PR to your review queue.  Thank you

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -360,18 +352,19 @@ public ConfigHolder generateConfigForSpecificNode(KafkaEndpoints kafkaEndpoints,
...
-        var clientEndpoint = kafkaEndpoints.getEndpointPair(Listener.EXTERNAL, nodeId);
-        var anonEndpoint = kafkaEndpoints.getEndpointPair(Listener.ANON, nodeId);
+    private ConfigHolder configureBroker(KafkaListenerSource kafkaListenerSource, ConfigHolder configHolder, TreeMap<String, String> protocolMap,
+                                         TreeMap<String, String> listeners,
+                                         TreeMap<String, String> advertisedListeners, TreeSet<String> earlyStart) {
+        var nodeId = configHolder.nodeId();
+        var nodeConfiguration = configHolder.properties();
+        var interBrokerEndpoint = kafkaListenerSource.getKafkaListener(Listener.INTERNAL, nodeId);
+        var clientEndpoint = kafkaListenerSource.getKafkaListener(Listener.EXTERNAL, nodeId);
+        var anonEndpoint = kafkaListenerSource.getKafkaListener(Listener.ANON, nodeId);
```
SamBarker, line 367: slightly confused by this as we seem to be resolving each listener but not actually using them.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -406,13 +398,13 @@ private boolean isAdditionalNode(int nodeId) {
...
 
-    private void configureTls(KafkaEndpoints.EndpointPair clientEndpoint, Properties server) {
+    private void configureTls(KafkaListener clientEndpoint, Properties server) {
         if (securityProtocol != null && securityProtocol.contains("SSL")) {
             if (brokerKeytoolCertificateGenerator == null) {
                 throw new RuntimeException("brokerKeytoolCertificateGenerator needs to be initialized when calling KafkaClusterConfig");
             }
             try {
-                brokerKeytoolCertificateGenerator.generateSelfSignedCertificateEntry("test@kroxylicious.io", clientEndpoint.getConnect().getHost(), "Dev",
+                brokerKeytoolCertificateGenerator.generateSelfSignedCertificateEntry("test@kroxylicious.io", clientEndpoint.advertised().host(), "Dev",
```
SamBarker, line 407: nit: variable name.

`clientEndpoint` -> `listener`?

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -481,60 +473,62 @@ private String deriveLoginModuleFromSasl(String saslMechanism) {
         }
     }
 
-    private static void configureInternalListener(TreeMap<String, String> protocolMap, TreeMap<String, String> listeners, KafkaEndpoints.EndpointPair interBrokerEndpoint,
+    private static void configureInternalListener(TreeMap<String, String> protocolMap, TreeMap<String, String> listeners, KafkaListener interBrokerEndpoint,
                                                   TreeMap<String, String> advertisedListeners, TreeSet<String> earlyStart, Properties server) {
         protocolMap.put(INTERNAL_LISTENER_NAME, SecurityProtocol.PLAINTEXT.name());
-        listeners.put(INTERNAL_LISTENER_NAME, interBrokerEndpoint.listenAddress());
-        advertisedListeners.put(INTERNAL_LISTENER_NAME, interBrokerEndpoint.advertisedAddress());
+        listeners.put(INTERNAL_LISTENER_NAME, interBrokerEndpoint.bind().toString());
```
SamBarker, line 479: nit: I don't mind the change from `listenAddress` to `bind` but needing to `toString` here feels like a step backwards.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -550,7 +544,7 @@ private static void putConfig(Properties server, String key, String value) {
      * Generates client connection config to connect to the anonymous listeners within the cluster. Thus bypassing all authentication mechanisms.
      *
      * @param bootstrapServers the bootstrap servers
-     * @return the anon connect config for cluster
+     * @return the anon kafkaNet config for cluster
```
SamBarker, line 547: `kafkaNet` ?

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -619,6 +613,13 @@ public Map<String, Object> getConnectConfigForCluster(String bootstrapServers, S
         return kafkaConfig;
     }
 
+    /**
+     * Generates client connection config to connect to the cluster via the supplied controller bootstrap address.
+     * This method is applicable to KRaft clusters only.
+     *
+     * @param bootstrapControllers controller bootstrap
+     * @return the connect config for cluster
+     */
```
SamBarker, line 622: should it throw in Zookeeper mode?

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaListener.java
```diff
@@ -0,0 +1,40 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * The type Kafka Listener.
+ *
+ * @param bind       the bind address
+ * @param kafkaNet   the connect address within the kafka network
+ * @param advertised the connect address from the client network
+ */
+public record KafkaListener(@NonNull KafkaEndpoint bind, @NonNull KafkaEndpoint kafkaNet, @NonNull KafkaEndpoint advertised) {
```
SamBarker, line 19: I think I'd prefer to see the `*Address` methods restored here to provide the correctly formatted address patterns.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -266,14 +271,14 @@ public String getBootstrapControllers() {
             throw new UnsupportedOperationException("Zookeeper based clusters don't support this operation");
         }
 
-        return buildBrokerList(nodeId -> getEndpointPair(Listener.CONTROLLER, nodeId));
+        return buildAdvertisedServers(this::isController, Listener.CONTROLLER);
     }
 
-    private synchronized String buildBrokerList(Function<Integer, KafkaClusterConfig.KafkaEndpoints.EndpointPair> endpointFunc) {
+    private synchronized String buildAdvertisedServers(Predicate<Integer> nodePredicate, Listener listener) {
```
SamBarker, line 277: does the predicate actually need to be passed in?

SamBarker (approved):
LGTM with nits

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: no, all three endpoint threes are used later in the method.   The naming is out of date.  I'll fix that.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: fixed.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: find/replace went rogue.  fixed!

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: It probably should, but KCC doesn't actually have state.  I don't really think the abstraction is quite right but the new code is in keeping with the existing code.  I'm not going to fix the deeper issue now.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: advertised servers can be the list of nodes in the KRaft role, or (currently) all nodes, so I think the predicate is important.

PR #455, opened 2025-03-17
[Experiment] Upgrade to Kafka 4.0.0 (RC4)

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md


k-wall:
Things to do:

* [ ] - Use reflective code for Zookeeper interactions
* [ ] - Separate Zookeeper based tests from KRaft ones so we can turn off tests if a version of Kafka that supports Zookeeper is not on the classpath.
* [ ] - Add matrix to Maven to run the tests the extension with older Kafka versions.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
```diff
@@ -180,8 +180,8 @@ void afterAll() {
 
     private static Stream<Version> versions() {
         return Stream.of(
-                version(Version.LATEST_RELEASE),
-                version(Version.LATEST_SNAPSHOT),
+                // version(Version.LATEST_RELEASE), TODO Remove once kafka native support 4.0.0 is available
```
k-wall, line 183: Revert this once the Kafka Native image is available

impl/src/main/java/io/kroxylicious/testing/kafka/clients/CloseableAdmin.java
```diff
@@ -292,19 +298,6 @@ public DescribeConfigsResult describeConfigs(Collection<ConfigResource> resource
         return instance.describeConfigs(resources, options);
     }
 
-    @SuppressWarnings("deprecation")
-    @Override
-    @Deprecated
-    public AlterConfigsResult alterConfigs(Map<ConfigResource, Config> configs) {
-        return instance.alterConfigs(configs);
```
k-wall, line 299: For backward compatibility sake we ought to maintain the method and use reflection to try to invoke the method on instance.  That way, we keep runtime compatibility for users with an older version of Kafka on the classpath who are using the deprecated methods.  My hunch is that the number of users doing this will be zero so I've going to avoid doing the work (& writing the tests).

franvila (approved):
LGTM

Issue #389, opened 2024-09-10
[User Poll] Migration to Kafka 4.0 -  removal of Zookeeper based Kafka clusters 


k-wall:
Kafka 4.0 is not too far away now so we need to think about how this project will transition to Kafka 4.0. 

This project currently has the ability to produce Kafka clusters for testing purposes.  It can produce either KRaft clusters or Kafka Clusters with Zookeeper.  The cluster can either run in-VM, or with Kafka Broker(s)/Zookeeper as containers.

The project realises its in-VM support using compile time dependencies on the Kafka libraries.  It literally creates 
instances of `kafka.server.KafkaServer` or `kafka.server.KafkaRaftServer` depending on your wishes.

When Kafka 4.0 comes along, inevitably some of the symbols this library depends upon will be gone when Kafka 4.x JARs are on the classpath.

It **might** be possible to retain the extension's support for Zookeeper based Kafka clusters when Kafka 3.x jars are on the classpath.   Reflective code might be one way or forcing a fallback onto the Container path might be another.    It is probably going to be a fair amount of work.

# Poll

We'd like to straw poll users of this library to gauge **the importance of ability this library to produce Zookeeper based clusters** (i.e. those clusters produced with the `@ZooKeeperCluster` annotation).

You might have code like this:

```java
    @Test
    void myFunkyTest(@BrokerCluster @ZooKeeperCluster KafkaCluster cluster)
            throws Exception {
        // do some stuff with cluster
    }
```

If we were to remove this feature, would this impact your test-cases?

Reply with emojis on to this comment:

😊 = content for Zookeeper based Kafka clusters support to be **removed**
🙁 = prefer Zookeeper  based Kafka clusters  support to continue.

More expansive thoughts welcomed too, just leave comments.

This poll is indicative.  We'll do our best to support the wishes of the majority but can't make a promise.

This poll with remain open until 2024-10-01.

k-wall:
In the end we decided that support will be retained for Zookeeper based clusters, when older versions of Kafka are on the classpath.  We've added a CI test matrix to make sure we stick to this commitment. #461

PR #456, opened 2025-03-18
Use reflection for all Zookeeper interactions

k-wall:
### Type of change

_Select the type of your PR_

- Refactoring

### Description

As a first step towards Kafka 4.0 support (and the retention of ability to form Kafka clusters with Zookeeper controllers when using older Kafka versions), refactor the code so that Zookeeper based Kafka is instantiated exclusively using reflection.

Also refactors the test suite so that test that need Zookeeper skip if Zookeeper is not present on the classpath.

This change does not **require** Kafka 4.0.0.  My intent is that this is committed first, then a follow up change will make the Kafka 4.0 specific changes.

Step towards #389 

### Additional Context

### Checklist

- [x] Write tests
- [x] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
@@ -9,19 +9,14 @@ package io.kroxylicious.testing.kafka.invm;
 import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
-import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Modifier;
 import java.net.InetSocketAddress;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -44,10 +39,7 @@ import org.apache.zookeeper.server.ZooKeeperServer;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import kafka.server.KafkaConfig;
 import kafka.server.KafkaRaftServer;
-import kafka.server.KafkaServer;
 import kafka.server.Server;
-import kafka.zk.AdminZkClient;
-import kafka.zk.KafkaZkClient;
 import scala.Option;
 
 import io.kroxylicious.testing.kafka.api.KafkaCluster;
@@ -59,8 +51,6 @@ import io.kroxylicious.testing.kafka.common.KafkaListenerSource;
 import io.kroxylicious.testing.kafka.common.PortAllocator;
 import io.kroxylicious.testing.kafka.common.Utils;
 import io.kroxylicious.testing.kafka.internal.AdminSource;
-
-import static kafka.zk.KafkaZkClient.createZkClient;
 
 /**
  * Configures and manages an in process (within the JVM) Kafka cluster.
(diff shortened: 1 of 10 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/invm/ReflectionUtils.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    private static boolean matchingMethod(Class<?>[] methodParameterTypes, Object... parameters) {
+        if (methodParameterTypes.length != parameters.length) {
+            return false;
+        }
+        boolean allMatch = true;
+        for (int i = 0; i < parameters.length; i++) {
+            // What about nulls?
```
SamBarker, line 90: I guess we have to assume `null` is a match. Or maybe we should throw here if we get a null passed through so we re-visit when we need it?

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -671,4 +699,14 @@ private void createClientCertificate() throws GeneralSecurityException, IOExcept
         this.clientKeytoolCertificateGenerator.generateSelfSignedCertificateEntry("clientTest@kroxylicious.io", "client", "Dev", "Kroxylicious.ip", null, null, "US");
     }
 
+    public static boolean zookeeperAvailable() {
```
SamBarker, line 702: nit: this duplicates the implementation in `AbstractExtensionTest` should `KafkaClusterTest` extend `AbstractExtensionTest`?

I guess `private static final boolean ZOOKEEPER_AVAILABLE = AbstractExtensionTest.zookeeperAvailable();` is the other option

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
```diff
@@ -124,10 +127,14 @@ void afterAll() {
 
     @SuppressWarnings("unused")
     static Stream<List<Annotation>> tuples() {
-        return Stream.of(
+        var tuples = new ArrayList<>(List.of(
                 List.of(brokerCluster(1), kraftCluster(1)),
-                List.of(brokerCluster(3), kraftCluster(1)),
-                List.of(brokerCluster(3), zooKeeperCluster()));
+                List.of(brokerCluster(3), kraftCluster(1))));
+        if (ZOOKEEPER_AVAILABLE) {
```
SamBarker, line 133: its an annoying pattern but I haven't thought of any brighter ideas.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -492,32 +484,59 @@ private List<String> buildScramArguments() {
...
+    }
+
+    private AutoCloseable createZkClient(String name, Time system, KafkaConfig config, ZKClientConfig zkClientConfig) {
+
+        try {
+            var clazz = Class.forName("kafka.zk.KafkaZkClient");
+            return ReflectionUtils.invokeStaticMethod(clazz, "createZkClient", name, system, config, zkClientConfig);
+        }
+        catch (Exception e) {
+            throw new IllegalStateException("Failed to invoke createZkClient while creating scram users", e);
```
SamBarker, line 525: is the exception message stale? I don't think this method is scram user specific.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -492,32 +484,59 @@ private List<String> buildScramArguments() {
...
 
+    private void changeConfigs(Object adminZkClient, String userEntityType, String name, Properties userConfig, boolean isUserClientId) {
+        ReflectionUtils.invokeInstanceMethod(adminZkClient, "changeConfigs", userEntityType, name, userConfig, isUserClientId);
+    }
+
+    private Properties fetchEntityConfig(Object adminZkClient, String userEntityType, String name) {
+        return ReflectionUtils.invokeInstanceMethod(adminZkClient, "fetchEntityConfig", userEntityType, name);
+    }
+
+    private AutoCloseable createZkClient(String name, Time system, KafkaConfig config, ZKClientConfig zkClientConfig) {
```
SamBarker, line 518: nit: do we really need to pass `name` and `time` down here? IntelliJ highlights that they are always invoked with the same args.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -121,45 +111,47 @@ private Server buildKafkaServer(KafkaClusterConfig.ConfigHolder c, List<String>
...
+    /**
+     * We instantiate the KafkaServer reflectively in order to continue to support older
+     * versions of Kafka that used Zookeeper based controllers.
+     *
+     * @param kafkaServerClazz kafka server class
+     * @param config           kafka config
+     * @param threadNamePrefix thread name prefix to be used by the server
+     */
+    @NonNull
+    private Server instantiateKafkaZookeeperServer(Class<Server> kafkaServerClazz, KafkaConfig config, Option<String> threadNamePrefix) {
```
SamBarker, line 140: Scala's option was a bit of surprise here. I see that its needed because it gets passed through as is to the construct call.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/ReflectionUtils.java
```diff
@@ -0,0 +1,96 @@
...
+                .filter(constructor -> Modifier.isPublic(constructor.getModifiers()))
+                .filter(constructor -> matchingMethod(constructor.getParameterTypes(), parameters))
+                .findFirst().map(constructor -> {
+                    try {
+                        return (K) constructor.newInstance(parameters);
+                    }
+                    catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+                        throw new IllegalStateException(e);
+                    }
+                });
```
SamBarker, line 51: extreme nit:
```suggestion
                .findFirst()
                .map(constructor -> {
                    try {
                        return (K) constructor.newInstance(parameters);
                    }
                    catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
                        throw new IllegalStateException(e);
                    }
                });
```

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
```diff
@@ -160,10 +167,13 @@ void testTuples(@ConstraintsMethodSource(value = "tuples", clazz = TemplateTest.
...
-            assertThat(observedTuples).isEqualTo(Set.of(
+            var expected = new HashSet<>(Set.of(
                     List.of(1, 1),
-                    List.of(3, 1),
-                    List.of(3, -1)));
+                    List.of(3, 1)));
+            if (ZOOKEEPER_AVAILABLE) {
+                expected.add(List.of(3, -1));
+            }
+            assertThat(observedTuples).isEqualTo(expected);
```
SamBarker, line 176: does `isEqualTo` care about the type of the set? Just wondering if this would be better expressed as 
```suggestion
            assertThat(observedTuples).containsExactlyInAnyOrderElementsOf(expected);
```

I wonder if that might be a bit more forgiving as I think this is one of the more temperamental tests

SamBarker (approved):
A few small things but over all good work @k-wall.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/ReflectionUtils.java
robobario, line 90: It looks like we don't expect to handle nulls anywhere, so maybe hoist the non-null parameter requirement to the top `invoke*` and `construct`.

Or we could change it to return true if it's non-primitive and parameter null. Then in `invokeMethod` collect to List, and throw if size != 1, so we only attempt the invoke if it's unambiguous.

SamBarker:
Very confused by why sonar thinks there is `0` coverage the test ran: https://github.com/kroxylicious/kroxylicious-junit5-extension/actions/runs/13930048481/job/38984625442#step:11:207

impl/src/main/java/io/kroxylicious/testing/kafka/invm/ReflectionUtils.java
k-wall: I think we should just reject any invocations with null args.   Like you say, we don't need any right now.  If we ever do, we can fix it then.

SamBarker:
It looks like we stopped getting coverage in November last year. I'm about to try a fix

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -492,32 +484,59 @@ private List<String> buildScramArguments() {
...
     // The accessibility hack is tactical for Kafka >=3.7.1 <= 4.0.0
     // The alternative is copying the code that constructs the ZK client see https://github.com/ozangunalp/kafka-native/pull/195/files as an example. Both options suck,
     // this is less code, and thus we hope less likely to bit rot.
     @SuppressWarnings("java:S3011")
-    private static void disableControllerCheck(KafkaZkClient zkClient) {
+    private static void disableControllerCheck(Object zkClient) {
         try {
-            final Class<? extends KafkaZkClient> zkClientClass = zkClient.getClass();
+            final Class<?> zkClientClass = zkClient.getClass();
             final Field enableEntityConfigControllerCheck = zkClientClass.getDeclaredField("enableEntityConfigControllerCheck");
```
robobario, line 540: I think we should shuffle the field manipulation to ReflectionUtils too

impl/src/main/java/io/kroxylicious/testing/kafka/invm/KraftLogDirUtil.java
```diff
@@ -48,6 +47,17 @@ static void prepareLogDirsForKraft(String clusterId, KafkaConfig config, List<St
         }
     }
 
+    private static MetadataVersion getMetadataVersion(KafkaConfig config) {
+        try {
+            var method = config.getClass().getMethod("interBrokerProtocolVersionString");
```
robobario, line 52: Call the ReflectionUtil instead

robobario (approved):
LGTM with nits

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/TemplateTest.java
k-wall: Agreed, I didn't much like the code, but it seemed like the only way.

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
k-wall: A neat abstraction isn't really leaping out to me.  We'd need the API to throw the NoSuchFieldException so we can cater for the case where the we are on <= 3.7.0.    I'll leave this for now

PR #461, opened 2025-03-19
Simple CI test matrix running the extension against the older Kafka versions

k-wall:
### Type of change

_Select the type of your PR_

- Enhancement / new feature

### Description

Simple CI test matrix running the extension against the older Kafka versions

Why: #455 we want the extension to keep working against older Kafka versions.  We therefore need adequate test support.

### Additional Context

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -46,6 +46,13 @@ jobs:
           distribution: 'zulu'
           cache: maven
 
+      - name: 'Cache Maven packages'
+        uses: actions/cache@v4
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+
       - name: Cache SonarCloud packages
         uses: actions/cache@v4
         with:
@@ -81,6 +88,58 @@ jobs:
         with:
           name: PR_NUMBER
           path: PR_NUMBER.txt
+
+      - name: Archive container logs
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: container-logs
+          path: "**/container-logs/**/*.log"
+
+  older_kafka_matrix:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        version: [3.8.0]
+    steps:
+      - name: 'Check out repository'
+        uses: actions/checkout@v4
+        with:
+          submodules: 'true'
+          fetch-depth: 0
+
+      - name: 'Set up Java'
+        uses: actions/setup-java@v4
+        with:
+          java-version: 17
+          distribution: 'zulu'
+          cache: maven
+
+      - name: 'Cache Maven packages'
+        uses: actions/cache@v4
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+
+      - name: 'Install junit extension'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: mvn -B install -DskipTests=true
+
+      - name: 'Run select junit tests (Kafka in-VM)'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # Note - the extension won't necessarily compile against the earlier versions of Kafka.
+        # We want to run the tests against the code that was compiled in the install step, just with
+        # a different version of Kafka on the classpath.
+        run: mvn -Dmaven.main.skip=true -Dmaven.javadoc.skip=true -pl impl,junit5-extension -B test -Dtest=KafkaClusterTest,ParameterExtensionTest -Dkafka.version=${{ matrix.version }}
+
+      - name: 'Run select junit tests (Kafka containers)'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: TEST_CLUSTER_EXECUTION_MODE=CONTAINER KAFKA_VERSION=${{ matrix.version }} mvn -Dmaven.main.skip=true -Dmaven.javadoc.skip=true -pl impl,junit5-extension -B test -Dtest=KafkaClusterTest,ParameterExtensionTest -Dkafka.version=${{ matrix.version }}
 
       - name: Archive container logs
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/maven.yaml
```diff
@@ -82,3 +89,55 @@ jobs:
...
+
+      - name: 'Install junit extension'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: mvn -B install -DskipTests=true
+
+      - name: 'Run select junit tests (Kafka in-VM)'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # Note - the extension won't necessarily compile against the earlier versions of Kafka.
```
k-wall, line 128: Maven doesn't make this easy.  Ideally you'd have project produce a test jar, then you'd point surefire at that artefact.  However, Maven doesn't allow such things.  I've banged my head for hours here before.

Instead, I'm running the tests against the installed artefacts.   The Junit5 test code needs to be compatible with older Kafka clients, otherwise we'd fail to compile.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -332,7 +332,9 @@ public ConfigHolder generateConfigForSpecificNode(KafkaListenerSource kafkaListe
         putConfig(nodeConfiguration, "listener.security.protocol.map", formatListeners(protocolMap, ":"));
         putConfig(nodeConfiguration, "listeners", formatListeners(listeners, "://"));
         putConfig(nodeConfiguration, "early.start.listeners", String.join(",", earlyStart));
-        putConfig(nodeConfiguration, "advertised.listeners", formatListeners(advertisedListeners, "://"));
+        if (!advertisedListeners.isEmpty()) {
```
k-wall, line 335: It turns out 3.8.0 in the KRaft controller case is sensitive to this parameter being present, even though it is empty.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -68,6 +70,12 @@
 @Timeout(value = 2, unit = TimeUnit.MINUTES)
 class KafkaClusterTest {
 
+    private static final System.Logger LOGGER = System.getLogger(KafkaClusterTest.class.getName());
+
+    static {
+        LOGGER.log(System.Logger.Level.INFO, "Kafka on classpath is version: {0}", AppInfoParser.getVersion());
```
k-wall, line 76: I just want to make it easy to see that the test matrix is doing the right thing.

.github/workflows/maven.yaml
```diff
@@ -82,3 +89,55 @@ jobs:
         with:
           name: container-logs
           path: "**/container-logs/**/*.log"
+
+  older_kafka_matrix:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        version: [3.8.0]
```
k-wall, line 98: We'll need to have a note somewhere to rev this.  How many previous version do we actually want to test with?

k-wall:
There is a test failure against 3.8.0 for test kraftAdminConnectionToControllers "Direct-to-controller communication is not supported with the current MetadataVersion." when running in container mode.  I haven't figured our why it is going wrong yet.  If I don't get to the bottom of the issue quickly,. i'll just raise an issue for it and add selectively disable that test for the 3.8.0/container combination.  I don't think it is worth getting side tracked.

.github/workflows/maven.yaml
SamBarker: I'm not sure we really need multiple older versions. Possibly just 3.9.0 (as the LTS IIRC with ZK support)?

.github/workflows/maven.yaml
SamBarker: I have seen hacks producing `-test` jars though I can't remember which plugin was used to generate the additional artefact.

.github/workflows/maven.yaml
```diff
@@ -82,3 +89,56 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: mvn -B install -DskipTests=true
+
+      - name: 'Run select junit tests (Kafka in-VM)'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # Note - the extension won't necessarily compile against the earlier versions of Kafka.
+        # We want to run the tests against the code that was compiled in the install step, just with
+        # a different version of Kafka on the classpath.
+        run: mvn -Dmaven.main.skip=true -Dmaven.javadoc.skip=true -pl impl,junit5-extension -B test -Dtest=KafkaClusterTest,ParameterExtensionTest -Dkafka.version=${{ matrix.version }}
```
SamBarker, line 131: Would it be easier if we added a `crossVersion` (better name required) profile to control the set of tests which are covered?

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
SamBarker: 😱

.github/workflows/maven.yaml
SamBarker: I can see an argument for testing `4.0` explicitly in the matrix as well in that it gives us confidence that things work across versions regardless of where main is.

SamBarker (approved):
A useful improvement @k-wall thanks for spending the time on it.

k-wall:
> There is a test failure against 3.8.0 for test kraftAdminConnectionToControllers "Direct-to-controller communication is not supported with the current MetadataVersion." when running in container mode. I haven't figured our why it is going wrong yet. If I don't get to the bottom of the issue quickly,. i'll just raise an issue for it and add selectively disable that test for the 3.8.0/container combination. I don't think it is worth getting side tracked.

I got the bottom of it!  The upshot is `kraftAdminConnectionToControllers` needs to be skipped when running CONTAINERS if the Kafka < 3.9.  This is because the client can't route to the controller until the work of https://issues.apache.org/jira/browse/KAFKA-16781.   I pushed a commit to conditionally skip the test.

.github/workflows/maven.yaml
k-wall: IIRC producing test jars is not too hard.  The diffcult was actually getting surefire to execute the tests from the JAR, rather than from compiled source files on a filesystem.

.github/workflows/maven.yaml
k-wall: Yes, I think that would be an improvement, but I think it can wait.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
k-wall: Test added.

PR #500, opened 2025-08-06
feat: junit extension prefers TEST_CLUSTER_EXECUTION_MODE env var when selecting a provisioning mechanism

robobario:
### Type of change

- Enhancement / new feature

### Description

This change introduces a way to control the selection of a cluster provider using an environment variable.

When multiple providers can satisfy the constraints for a cluster injection, you can now set the `TEST_CLUSTER_PREFERRED_EXECUTION_MODE` environment variable to prefer one of the modes. If the variable is not set, the existing logic of preferring the fastest provider remains unchanged.

This is necessary because the current behavior defaults to the fastest provider, making it impossible to force an entire test suite to prefer Testcontainers. I would like to be able to run the kroxylicious/kroxylicious tests, preferring to use testcontainers wherever possible, to exercise container mode.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
@@ -28,6 +28,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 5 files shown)


junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
```diff
@@ -1182,6 +1191,23 @@ static KafkaClusterProvisioningStrategy findBestProvisioningStrategy(
                 });
     }
 
+    // if the user sets TEST_CLUSTER_PREFERRED_EXECUTION_MODE, we want to prefer this mode if the user declares a generic KafkaCluster
+    private static Function<KafkaClusterProvisioningStrategy, Integer> preferStrategiesMatchingEnvVar() {
+        return x -> {
+            String preferredModeFromEnv = System.getenv().get(TEST_CLUSTER_PREFERRED_EXECUTION_MODE);
```
k-wall, line 1197: The env var TEST_CLUSTER_EXECUTION_MODE is used by KafkaClusterFactory code path (which is not used by the KafkaClusterProvisioningStrategy).   I wondering if we should use the same environment variable rather than introducing a new one.

k-wall (approved):

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtension.java
robobario: sounds good, I think sharing the env var will be clear enough despite the subtleties

robobario: going to add some docs to the front README with details about how to influence the provisioning mechanism for users.

Issue #364, opened 2024-07-24
[KIP-974] Consider integration Apache Kafka native images.


k-wall:
Currently the test extension uses Quarkus based kafka-native images from xxx.  Kroxylicious-junit5-ext should consider whether the work of KIP-974 https://github.com/apache/kafka/blob/trunk/docker/native/README.md should be made available, either instead of, or in addition to the current support.

k-wall:
I think we can make this largely transparent to the user of the extension.  If the user is using a KRaft Kafka with a version >= 3.8.0, the test framework could default to using Apache based images.  Otherwise, the behaviour needs to be at today.  It should be possible to override the image location in either case.

robobario:
An update on this, we want to switch to apache for 4.1 (I believe we aren't going to release any more versions of ozangunalp/kafka-native?)

## SASL not working on apache native image
I found a couple of issues with the upstream native image around SASL and classes that can't be reflectively loaded.

https://issues.apache.org/jira/browse/KAFKA-19583
https://issues.apache.org/jira/browse/KAFKA-19584

 Fixes for these aren't likely to go into the 4.1 release this late, so we might hope to see them in a later point release.

I think we should implement it so that we prefer apache-native for 4.0+, and if SASL features are enabled we will use the non-native image. Plus give users an annotation to influence whether we run native or not. Like:

```
@ExecutionVariant(Variant.NATIVE | Variant.JAVA)
```

With rules:
1. InVM defaults to a JAVA variant and cannot run a NATIVE variant (so in junit testcontainers should be selected)
2. Testcontainers defaults to a NATIVE variant, unless the user configures a sasl mechanism, in which case we default to JAVA
3. If the user explicitly specifies `@ExecutionVariant(Variant.NATIVE)` and a sasl mechanism, we fail fast.

Alternatively we could make it explicitly about containers, like, `@ImageFlavour(NATIVE | JAVA)` and the InVM mode wouldn't support that annotation.

## apache images dont have SCRAM user convenience

Another difference is that the old image takes an env var where you can pass SCRAM users, and the image knows how to use them when formatting the storage. The apache images do not have this convenience.

Instead, we can start up the broker and then using an admin client connected to the anonymous listener, create the scram credentials.

robobario:
FYI we decided to depend on the JRE-based upstream apache images, lets leave this issue open though and we can revisit in future

PR #501, opened 2025-08-06
Kafka 4.1.0 upgrade

robobario:
### Type of change

- Enhancement / new feature

### Description

Upgrade to kafka 4.1.0

Adopts upstream apache JRE-based Kafka images for 4.1.0+ for Testcontainers deployment mode

A notable difference is that we now use the Admin client to create the SASL SCRAM users for both in-vm and testcontainer mode. This avoids having to use internal code when formatting the storage, using a public API to install the users.

Currently we cannot use the native variant of the apache image as there are some problems that prevent the use of SASL authentication.
https://issues.apache.org/jira/browse/KAFKA-19583
https://issues.apache.org/jira/browse/KAFKA-19584

If your test targets older versions of Kafka <4.1.0 then they will use the established `quay.io/ogunalp/kafka-native`.

The apache images are pulled from `mirror.gcr.io/apache/kafka`, this can be overridden with a new environment variable `APACHE_KAFKA_IMAGE_REPO`. The existing `KAFKA_IMAGE_REPO` environment variable will only be applied to older target kafka version using images from `quay.io/ogunalp/kafka-native`.

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
@@ -85,13 +85,16 @@ import static io.kroxylicious.testing.kafka.common.Utils.awaitExpectedBrokerCoun
  */
 public class TestcontainersKafkaCluster implements Startable, KafkaCluster, KafkaListenerSource, AdminSource {
 
-    public static final String ZOOKEEPER_IMAGE_TAG = "ZOOKEEPER_IMAGE_TAG";
-    public static final String ZOOKEEPER_IMAGE_REPO = "ZOOKEEPER_IMAGE_REPO";
+    @Deprecated(since = "0.12.0")
+    public static final String LEGACY_ZOOKEEPER_IMAGE_TAG = "ZOOKEEPER_IMAGE_TAG";
+    @Deprecated(since = "0.12.0")
+    public static final String LEGACY_ZOOKEEPER_IMAGE_REPO = "ZOOKEEPER_IMAGE_REPO";
     private static final System.Logger LOGGER = System.getLogger(TestcontainersKafkaCluster.class.getName());
     /**
-     * environment variable specifying the native kafka image repository.
+     * environment variable specifying the ogunalp native kafka image repository.
      */
-    public static final String KAFKA_IMAGE_REPO = "KAFKA_IMAGE_REPO";
+    @Deprecated(since = "0.12.0")
+    public static final String LEGACY_KAFKA_IMAGE_REPO = "KAFKA_IMAGE_REPO";
     /**
      * environment variable specifying the apache kafka image repository.
      */
@@ -114,10 +117,10 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
     private static final int ZOOKEEPER_PORT = 2181;
 
     @Deprecated(since = "0.12.0")
-    private static final String QUAY_KAFKA_IMAGE_REPO = "quay.io/ogunalp/kafka-native";
+    private static final String LEGACY_QUAY_KAFKA_IMAGE_REPO = "quay.io/ogunalp/kafka-native";
     private static final String APACHE_KAFKA_JAVA_IMAGE_REPO = "mirror.gcr.io/apache/kafka";
     @Deprecated(since = "0.12.0")
-    private static final String QUAY_ZOOKEEPER_IMAGE_REPO = "quay.io/ogunalp/zookeeper-native";
+    private static final String LEGACY_QUAY_ZOOKEEPER_IMAGE_REPO = "quay.io/ogunalp/zookeeper-native";
     private static final int CONTAINER_STARTUP_ATTEMPTS = 3;
     private static final Duration STARTUP_TIMEOUT = Duration.ofMinutes(2);
     private static final Duration RESTART_BACKOFF_DELAY = Duration.ofMillis(2500);
@@ -133,7 +136,7 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
 
     // This uid needs to match the uid used by the ozangunalp native kafka container to execute the kafka process
     @Deprecated(since = "0.12.0")
-    private static final String KAFKA_CONTAINER_UID = "1001";
+    private static final String LEGACY_KAFKA_CONTAINER_UID = "1001";
     private static final int READY_TIMEOUT_SECONDS = 120;
     private static final String LOCALHOST = "localhost";
     private static final Pattern MAJOR_MINOR_PATCH = Pattern.compile("\\d+(\\.\\d+(\\.\\d+)?)?");
@@ -198,7 +201,8 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
             this.zookeeper = null;
         }
         else {
-            zookeeperImage = resolveImage(TestcontainersKafkaCluster::zookeeperRegistryResolver, () -> envVarValueOrElse(ZOOKEEPER_IMAGE_TAG, Version.LATEST_RELEASE));
+            zookeeperImage = resolveImage(TestcontainersKafkaCluster::zookeeperRegistryResolver,
+                    () -> envVarValueOrElse(LEGACY_ZOOKEEPER_IMAGE_TAG, Version.LATEST_RELEASE));
             this.zookeeper = new ZookeeperContainer(zookeeperImage)
                     .withName(name)
                     .withNetwork(network)
@@ -363,7 +367,7 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
             return DockerImageName.parse(envVarValueOrElse(APACHE_KAFKA_IMAGE_REPO, APACHE_KAFKA_JAVA_IMAGE_REPO));
         }
         else {
-            return DockerImageName.parse(envVarValueOrElse(KAFKA_IMAGE_REPO, QUAY_KAFKA_IMAGE_REPO));
+            return DockerImageName.parse(envVarValueOrElse(LEGACY_KAFKA_IMAGE_REPO, LEGACY_QUAY_KAFKA_IMAGE_REPO));
         }
     }
 
@@ -390,7 +394,7 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
     }
 
     private static @NonNull DockerImageName zookeeperRegistryResolver() {
-        return DockerImageName.parse(envVarValueOrElse(ZOOKEEPER_IMAGE_REPO, QUAY_ZOOKEEPER_IMAGE_REPO));
+        return DockerImageName.parse(envVarValueOrElse(LEGACY_ZOOKEEPER_IMAGE_REPO, LEGACY_QUAY_ZOOKEEPER_IMAGE_REPO));
     }
 
     /**
@@ -727,7 +731,7 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
     @SuppressWarnings({ "try" })
     private static String createNamedVolume(KafkaClusterConfig clusterConfig) {
         try (DockerClient dockerClient = createDockerClient(); var volumeCmd = dockerClient.createVolumeCmd();) {
-            String containerUid = clusterConfig.isKafkaVersion41OrHigher() ? APACHE_CONTAINER_UID : KAFKA_CONTAINER_UID;
+            String containerUid = clusterConfig.isKafkaVersion41OrHigher() ? APACHE_CONTAINER_UID : LEGACY_KAFKA_CONTAINER_UID;
             if (CONTAINER_ENGINE_PODMAN) {
                 volumeCmd.withDriverOpts(Map.of("o", "uid=" + containerUid));
             }
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactory.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterFactory.java
@@ -32,10 +32,10 @@ public class KafkaClusterFactory {
      * @deprecated use the field from {@link TestcontainersKafkaCluster}
      */
     @Deprecated(forRemoval = true)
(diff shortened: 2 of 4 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/clients/CloseableAdmin.java
```diff
@@ -14,116 +14,7 @@
...
-import org.apache.kafka.clients.admin.RemoveRaftVoterOptions;
-import org.apache.kafka.clients.admin.RemoveRaftVoterResult;
-import org.apache.kafka.clients.admin.RenewDelegationTokenOptions;
-import org.apache.kafka.clients.admin.RenewDelegationTokenResult;
-import org.apache.kafka.clients.admin.UnregisterBrokerOptions;
-import org.apache.kafka.clients.admin.UnregisterBrokerResult;
-import org.apache.kafka.clients.admin.UpdateFeaturesOptions;
-import org.apache.kafka.clients.admin.UpdateFeaturesResult;
-import org.apache.kafka.clients.admin.UserScramCredentialAlteration;
+import org.apache.kafka.clients.admin.*;
```
robobario, line 17: bad IDEA

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -250,16 +254,27 @@ private KafkaContainer buildKafkaContainer(KafkaClusterConfig.ConfigHolder holde
...
+                .withCopyToContainer(Transferable.of(propertiesToBytes(properties), 0644), "/etc/kafka/docker/server.properties")
                 .withStartupAttempts(CONTAINER_STARTUP_ATTEMPTS)
                 .withMinimumRunningDuration(MINIMUM_RUNNING_DURATION)
                 .withStartupTimeout(STARTUP_TIMEOUT);
 
         if (clusterConfig.isSaslScram() && !clusterConfig.getUsers().isEmpty()) {
+            if (clusterConfig.isKafkaPostVersion4()) {
+                throw new RuntimeException("scram not implemented for apache kafka");
+            }
+            // todo get scram users into apache native
```
robobario, line 277: wondering, do we need to support the ozan-native-style configuration where we format the storage with scram users, so that nodes can uses scram auth for intercommunication? Should we instead use the admin client to create the scram creds after the cluster is running?

robobario:
mm tests failing with what might be an upstream issue, native image failing to reflectively call the no-args constructor OAuthBearerUnsecuredLoginCallbackHandler. Wonder if it needs to be added to some graal config.


```
2025-08-06 05:14:09 ERROR testcontainers-lifecycle-0 tc.docker.io/apache/kafka-native:latest:549 - Log output from the failed container:
===> User
uid=1000(appuser) gid=1000(appuser) groups=1000(appuser)
===> Setting default values of environment variables if not already set.
===> Configuring ...
===> Launching ... 
WARNING: THIS IS AN EXPERIMENTAL DOCKER IMAGE RECOMMENDED FOR LOCAL TESTING AND DEVELOPMENT PURPOSES.
[...
[2025-08-06 05:14:08,765] ERROR [BrokerServer id=0] Fatal error during broker startup. Prepare to shutdown (kafka.server.BrokerServer)
org.apache.kafka.common.KafkaException: org.apache.kafka.common.KafkaException: Could not find a public no-argument constructor for org.apache.kafka.common.security.oauthbearer.internals.unsecured.OAuthBearerUnsecuredLoginCallbackHandler
	at org.apache.kafka.common.network.SaslChannelBuilder.configure(SaslChannelBuilder.java:184) ~[?:?]
	at org.apache.kafka.common.network.ChannelBuilders.create(ChannelBuilders.java:188) ~[?:?]
	at org.apache.kafka.common.network.ChannelBuilders.serverChannelBuilder(ChannelBuilders.java:105) ~[?:?]
	at kafka.network.Processor.<init>(SocketServer.scala:883) ~[?:?]
	at kafka.network.Acceptor.newProcessor(SocketServer.scala:791) ~[kafka.Kafka:?]
	at kafka.network.Acceptor.$anonfun$addProcessors$1(SocketServer.scala:757) ~[kafka.Kafka:?]
	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:190) ~[kafka.Kafka:?]
	at kafka.network.Acceptor.addProcessors(SocketServer.scala:756) ~[kafka.Kafka:?]
	at kafka.network.DataPlaneAcceptor.configure(SocketServer.scala:472) ~[?:?]
	at kafka.network.SocketServer.createDataPlaneAcceptorAndProcessors(SocketServer.scala:222) ~[?:?]
	at kafka.network.SocketServer.$anonfun$new$16(SocketServer.scala:149) ~[?:?]
	at kafka.network.SocketServer.$anonfun$new$16$adapted(SocketServer.scala:149) ~[?:?]
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619) ~[kafka.Kafka:?]
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617) ~[kafka.Kafka:?]
	at scala.collection.AbstractIterable.foreach(Iterable.scala:935) ~[kafka.Kafka:?]
	at kafka.network.SocketServer.<init>(SocketServer.scala:149) ~[?:?]
	at kafka.server.BrokerServer.startup(BrokerServer.scala:274) ~[?:?]
	at kafka.server.KafkaRaftServer.$anonfun$startup$2(KafkaRaftServer.scala:96) ~[?:?]
	at kafka.server.KafkaRaftServer.$anonfun$startup$2$adapted(KafkaRaftServer.scala:96) ~[?:?]
	at scala.Option.foreach(Option.scala:437) [kafka.Kafka:?]
	at kafka.server.KafkaRaftServer.startup(KafkaRaftServer.scala:96) [kafka.Kafka:?]
	at kafka.Kafka$.main(Kafka.scala:97) [kafka.Kafka:?]
	at kafka.docker.KafkaDockerWrapper$.main(KafkaDockerWrapper.scala:68) [kafka.Kafka:?]
	at kafka.docker.KafkaDockerWrapper.main(KafkaDockerWrapper.scala) [kafka.Kafka:?]
	at java.base/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH) [kafka.Kafka:?]
Caused by: org.apache.kafka.common.KafkaException: Could not find a public no-argument constructor for org.apache.kafka.common.security.oauthbearer.internals.unsecured.OAuthBearerUnsecuredLoginCallbackHandler
	at org.apache.kafka.common.utils.Utils.newInstance(Utils.java:403) ~[?:?]
	at org.apache.kafka.common.security.authenticator.LoginManager.<init>(LoginManager.java:65) ~[?:?]
	at org.apache.kafka.common.security.authenticator.LoginManager.acquireLoginManager(LoginManager.java:116) ~[?:?]
	at org.apache.kafka.common.network.SaslChannelBuilder.configure(SaslChannelBuilder.java:170) ~[?:?]
	... 24 more
Caused by: java.lang.NoSuchMethodException: org.apache.kafka.common.security.oauthbearer.internals.unsecured.OAuthBearerUnsecuredLoginCallbackHandler.<init>()
	at java.base/java.lang.Class.checkMethod(DynamicHub.java:1075) ~[kafka.Kafka:?]
	at java.base/java.lang.Class.getConstructor0(DynamicHub.java:1238) ~[kafka.Kafka:?]
	at java.base/java.lang.Class.getDeclaredConstructor(DynamicHub.java:2930) ~[kafka.Kafka:?]
	at org.apache.kafka.common.utils.Utils.newInstance(Utils.java:401) ~[?:?]
	at org.apache.kafka.common.security.authenticator.LoginManager.<init>(LoginManager.java:65) ~[?:?]
	at org.apache.kafka.common.security.authenticator.LoginManager.acquireLoginManager(LoginManager.java:116) ~[?:?]
	at org.apache.kafka.common.network.SaslChannelBuilder.configure(SaslChannelBuilder.java:170) ~[?:?]
	... 24 more
```

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
k-wall: > Should we instead use the admin client to create the scram creds after the cluster is running?

I think so, let's use the Kafka public APIs everywhere.

k-wall:
> I tried out share groups and found it didn't work neatly out of the box,
> by default the internal topic tries to be configured with replication factor 3 and fails with our single-node default cluster.

You are talking about just "__share_group_state", right?  I think they'll be edge cases associated with add/remove node.   IIRC we had to tune them for those to suit those use-cases.


> I think we should do the same in in-vm mode to make it convenient to test with Share Groups

Why would we have different behaviour for in-VM vs Container?

General point, I'm not against getting the extension to play nicely with the Share Groups and it would be certainly be nice to ready when the Share Groups feature goes GA in Kafka.  But I think a separate issue is the way to go and I don't think it is the highest priority thing right now.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -80,6 +81,7 @@ public class KafkaClusterConfig {
     private static final String SCRAM_256_SASL_MECHANISM_NAME = "SCRAM-SHA-256";
     private static final String SCRAM_512_SASL_MECHANISM_NAME = "SCRAM-SHA-512";
     private static final Pattern PRE_KAFKA_39 = Pattern.compile("^3\\.[0-8]\\..*$");
+    private static final Pattern POST_KAFKA_4_1 = Pattern.compile("^4\\.[1-9][0-9]*\\..*$");
```
k-wall, line 84: Post 4.1 to means means 4.2 onwards.  Do you mean post 4.0?

robobario: Maybe inverting it to be like the existing 'pre-3.9' language would be less ambiguous.  So we would apply the old logic to pre-4.1

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -407,6 +432,11 @@ public synchronized void start() {
                     clusterConfig.getAnonConnectConfigForCluster(buildBrokerListFor(Listener.ANON)),
                     READY_TIMEOUT_SECONDS, TimeUnit.SECONDS,
                     clusterConfig.getBrokersNum());
+            if (clusterConfig.isSaslScram() && !clusterConfig.getUsers().isEmpty()) {
+                if (clusterConfig.isKafkaPostVersion41()) {
```
k-wall, line 436: Could we switch to this approach wholesale (i mean <4.0)?   Would it be much work. The motivation to to have fewer conditions.

robobario:
> > I tried out share groups and found it didn't work neatly out of the box,
> > by default the internal topic tries to be configured with replication factor 3 and fails with our single-node default cluster.
> 
> You are talking about just "__share_group_state", right? I think they'll be edge cases associated with add/remove node. IIRC we had to tune them for those to suit those use-cases.
> 
> > I think we should do the same in in-vm mode to make it convenient to test with Share Groups
> 
> Why would we have different behaviour for in-VM vs Container?

My  mistake, we control the full set of properties for both strategies, I thought the built in defaults of the container were involved.

> General point, I'm not against getting the extension to play nicely with the Share Groups and it would be certainly be nice to ready when the Share Groups feature goes GA in Kafka. But I think a separate issue is the way to go and I don't think it is the highest priority thing right now.

Will make an issue, I just saw this issue for adding a closeable share consumer and thought it would be an easy thing to include in the 4.1 branch, but testing it I couldn't get it to consume any messages, took quite a while to spot this topic creation failure as you need logs set to DEBUG

impl/src/main/java/io/kroxylicious/testing/kafka/invm/InVMKafkaCluster.java
```diff
@@ -463,78 +456,6 @@ private static void trapKafkaSystemExit() {
         return CloseableAdmin.create(clusterConfig.getAnonConnectConfigForCluster(buildBrokersListFor(Listener.ANON)));
     }
 
-    private List<String> buildScramArguments() {
```
k-wall, line 466: Glad to see all of this die

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -109,6 +112,7 @@ public class TestcontainersKafkaCluster implements Startable, KafkaCluster, Kafk
     private static final int ZOOKEEPER_PORT = 2181;
 
     private static final String QUAY_KAFKA_IMAGE_REPO = "quay.io/ogunalp/kafka-native";
+    private static final String APACHE_KAFKA_JAVA_IMAGE_REPO = "docker.io/apache/kafka";
```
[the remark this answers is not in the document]
robobario: I went with using mirror.gcr.io as default, plus an env var `APACHE_KAFKA_IMAGE_REPO` allowing users to override, this follows the existing mechanisms.

I did wonder a little whether we should repurpose `KAFKA_IMAGE_REPO`. Like add a `LEGACY_KAFKA_IMAGE_REPO` variable and make a little hierarchy of preference like:

kafka <4.1: userDefinedVersion > LEGACY_KAFKA_IMAGE_REPO > KAFKA_IMAGE_REPO > default
kafka >=4.1: userDefinedVersion > APACHE_KAFKA_IMAGE_REPO > KAFKA_IMAGE_REPO > default

so that a user who just cares about the latest kafka version could use KAFKA_IMAGE_REPO, but I figure it's a bit much.

On that, maybe we should have a table of env vars on the front README.

robobario:
I updated #364 to note that we will use the JRE images, we can keep that issue open to revisit the upstream native images in future

Also created a ticket to make testing Kafka Queues more convenient when it goes GA #510

robobario:
and a successful run using container mode:

https://github.com/robobario/kroxylicious/actions/runs/17571164043/job/49907510188?pr=36

a bit of tweaking needed to fix the image mirror substitutor to not replace the `OneShotContainer` image from `*.redhat.com` (should we really be using `latest`?) and also a test needed an await added, I suspect just a little slower for a transaction to commit with the image version.

PR #519, opened 2025-09-15
Enable users to customize the names of topics created by the extension

robobario:
### Type of change

- Enhancement / new feature

### Description

Users can annotate a Topic parameter or field with an `@TopicNameMethodSource` to customize the topic naming.

Following the other method source annotations users can supply:
1. `clazz` which points at the class carrying the method to invoke. The
   default value is Void which implies that we target the Test class.
2. `value` which is the name of the method. It must have no arguments
   and return String. It must be static.

Then, when the extension is automatically creating topics the user has requested, this method will be invoked to retrieve the topic name.

This enables users to control the topic naming for cases where the length and characters in the topic name have some restrictions differing from the default automatic naming convention.

The immediate driver is that in kroxylicious/kroxylicious we want to arrange things such that the topic names created for some tests also happen to be valid Key Names in Azure Key Vault.

This replaces #517 

### Checklist

- [ ] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/Topic.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/Topic.java
@@ -13,7 +13,8 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
  * Signals the wish for a topic to be created and injected into a test as either test parameter
- * or field.
+ * or field. To customize the naming of the topic, parameters and fields of this type should be
+ * annotated with {@link TopicNameMethodSource}.
  * {@see KafkaClusterExtension}
  */
 public interface Topic {
--- a/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicNameMethodSource.java
+++ b/junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicNameMethodSource.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Kroxylicious Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.kroxylicious.testing.kafka.junit5ext;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * {@link TopicNameMethodSource} is used to customize topic naming. It may be applied to the {@link Topic} type.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.PARAMETER })
+public @interface TopicNameMethodSource {
+
+    /**
+     * The name of the {@code static}, package- or {@code public}-accessible method.
+     * Defaults to 'topicName'.
+     * @return the string
+     */
+    String value() default "topicName";
+
+    /**
+     * The class where defining the static method, or Void.class (default), if the method
+     * is defined with the class defining the annotation test.
+     * @return the class
+     */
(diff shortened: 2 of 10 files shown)


junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/Topic.java
```diff
@@ -13,7 +13,7 @@
 
 /**
  * Signals the wish for a topic to be created and injected into a test as either test parameter
- * or field.
+ * or field. The name will conform to the pattern '[a-z]+_[a-z]+'.
```
robobario, line 16: wondering if this is useful, maybe it should just say that tests should assume that any characters that are allowed by Kafka could be used in the name and have a link off to the annotation saying "if you need to control the naming, use this annotation".

k-wall: I'd mention the annotation as you suggest.

junit5-extension/src/main/java/io/kroxylicious/testing/kafka/junit5ext/TopicNameMethodSource.java
```diff
@@ -0,0 +1,34 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.PARAMETER })
+public @interface TopicNameMethodSource {
+
+    /**
+     * The name of the {@code static}, package- or {@code public}-accessible method.
+     * Defaults to 'topicName'.
+     * @return the string
+     */
+    String value() default "topicName";
```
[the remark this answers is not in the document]
k-wall: Ah - I see you've done what DimensionMethodSource does.  Withdrawing comment.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/AbstractExtensionTest.java
```diff
@@ -23,6 +23,9 @@
 import static org.assertj.core.api.InstanceOfAssertFactories.STRING;
 
 public abstract class AbstractExtensionTest {
+    public static final String TOPIC_NAME_PATTERN = "^[a-z]+_[a-z]+$";
```
k-wall, line 26: I'm not sure the TOPIC_NAME_PATTERN constant actually helps understandability.  The pattern is used to test the value returned by the Moby Names Generator.  Calling it TOPIC_NAME_PATTERN makes it look like it used more widely.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/KafkaClusterExtensionTest.java
```diff
@@ -0,0 +1,189 @@
...
+
+    public static Stream<Arguments> testClassAndDefault() {
+        return Stream.of(Arguments.argumentSet("explicit", KafkaClusterExtensionTest.class), Arguments.argumentSet("default", Void.class));
+    }
+
+    @Test
+    void randomHumanReadableName() {
+        RandomHumanReadable randomHumanReadable = new RandomHumanReadable();
+        String name = randomHumanReadable.name();
+        assertThat(name).isNotNull().matches(TOPIC_NAME_PATTERN);
```
k-wall, line 48: Could use Kafka's Topic.validate() here.

junit5-extension/src/test/java/io/kroxylicious/testing/kafka/junit5ext/AbstractExtensionTest.java
robobario: yeah, we've got plenty of coverage showing that the names result in topics being created. I'll drop the pattern checks.

PR #537, opened 2025-10-28
add IP support for SAN when the domain is an IP

franvila:
### Type of change

- Enhancement

### Description

In case we want to create a certificate that accepts IP addresses as a domain, we should add the SAN with the IP prefix.

### Additional Context

For those cases where we cannot have a DNS, such as accessing to a kubernetes cluster on minikube, we would need the capacity of doing it using the IP:Port, so we need the SAN in the certificate to accept also that IP address.

### Checklist

- [x] Write tests
- [x] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGenerator.java
@@ -21,8 +21,9 @@ import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.bouncycastle.util.IPAddress;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import static java.lang.System.Logger.Level.DEBUG;
 import static java.lang.System.Logger.Level.WARNING;


impl/src/test/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGeneratorTest.java
```diff
@@ -40,6 +41,30 @@ public void generatesKeyStore() throws Exception {
...
+
+        var ks = KeyStore.getInstance(keystoreFile, password.toCharArray());
+        var aliases = aliasList(ks);
+        assertThat(aliases).hasSize(1);
+        var alias = aliases.get(0);
+        assertThat(ks.getCertificate(alias)).isNotNull();
+        assertThat(ks.getKey(alias, password.toCharArray())).isNotNull();
+        assertThat(ks.getType()).isEqualTo(generator.getKeyStoreType());
+        // 7 corresponds to the IPAddress SAN type
+        assertThat(((X509Certificate) ks.getCertificate(alias)).getSubjectAlternativeNames().stream().findFirst().filter(a -> (int) a.get(0) == 7).map(a -> a.get(1)))
```
k-wall: this is a bit impenetrable.  I'd introduce an intermediate variable or two

impl/src/test/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGeneratorTest.java
```diff
@@ -40,6 +41,30 @@ public void generatesKeyStore() throws Exception {
...
+                "Kroxylicious.io", null, null, "US");
+
+        var keystore = generator.getKeyStoreLocation();
+        assertThat(keystore).isNotEmpty();
+        var keystoreFile = new File(keystore);
+        assertThat(keystoreFile).exists();
+        var password = generator.getPassword();
+
+        var ks = KeyStore.getInstance(keystoreFile, password.toCharArray());
+        var aliases = aliasList(ks);
```
k-wall: Ideally, tests should test one thing.  This test is all about SAN IP address - so just test that aspect.

k-wall (approved):
couple of nits. but otherwise LGTM

PR #544, opened 2025-11-03
replace KeyToolCertificateGenerator by CertificateBuilder from netty

### LINKED ISSUE #131: KeytoolCertificateGenerator: Improve API and implementation
`KeytoolCertificateGenerator` was a great start that has allowed us to quickly get testing TLS use-cases.  However, its API is a [bit weird](https://github.com/kroxylicious/kroxylicious-junit5-extension/blob/5c5501d5d4bd5bdd566be2e3d4d537ffaa8e218f/impl/src/test/java/io/kroxylicious/testing/kafka/common/KeytoolCertificateGeneratorTest.java#L50)  and it relies on shelling out to command line tools.

We should think about:

* Do we value the integration tests having the ability to generate TLS certificates on the fly?  Maybe static key material would be good enough.
* Shelling out to command line tools makes the test suite slower, introduces fragility and erects barriers to participation.  Developers need to keep their environments in sync with the same version of the command line tools.  This isn't really reasonable for open source development.
* Self-signed certificate generation in Java isn't possible using the public APIs.  We could adopt Bouncy Castle as a test dependency.  We could then take an approach similar to Strimzi https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/main/java/io/strimzi/systemtest/security/SystemTestCertAndKeyBuilder.java

Also - as originally [discussed on #120](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/120), it would be desirable if the generator supported other keystore formats, and the generated key material was available over an API to support the testing of richer TLS use-cases.  It would also be good if it supported things like generating expired certificates, and certificate chains.

franvila:
### Type of change

- Enhancement / new feature

### Description

Replace `KeyToolCertificateGenerator` by `KeystoreManager` using netty `CertificateBuilder` class.

### Additional Context

Closes #131 

### Checklist

- [x] Write tests
- [ ] Update documentation
- [ ] Update CHANGELOG.md

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KeystoreManager.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KeystoreManager.java
@@ -17,13 +17,10 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
-import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-
-import org.bouncycastle.util.IPAddress;
 
 import io.netty.pkitesting.CertificateBuilder;
 import io.netty.pkitesting.X509Bundle;
@@ -42,23 +39,6 @@ public class KeystoreManager {
                 .rsa2048()
                 .subject(distinguishedName);
     }
-
-//    /**
-//     * Add Subject Alternative Names (SAN) to a certificate builder.
-//     *
-//     * @param certificateBuilder the certificate builder
-//     * @param sanNames the names for SAN: They can be DNS names and/or IP Addresses
-//     */
-//    public void addSanNames(CertificateBuilder certificateBuilder, List<String> sanNames) {
-//        sanNames.forEach(name -> {
-//            if (IPAddress.isValidIPv4(name) || IPAddress.isValidIPv6(name)) {
-//                certificateBuilder.addSanIpAddress(name);
-//            }
-//            else {
-//                certificateBuilder.addSanDnsName(name);
-//            }
-//        });
-//    }
 
     /**
      * Builds and adds provided certificate builder as a self-signed certificate


impl/src/main/java/io/kroxylicious/testing/kafka/common/CertificateGenerator.java
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * @param domain the domain
+     * @param ipAddress the ip address
+     */
+    public CertificateGenerator(String domain, String ipAddress) {
+        this.domain = domain;
+        this.ipAddress = ipAddress;
+    }
+
+    private String getDomainName(String email, String organizationUnit, String organization, String city, String state, String country) {
```
SamBarker: The PKI term i, I think,  `Distinguished Name`, see [rfc5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4) so this should be renamed 
```suggestion
    private String getDistinguishedName(String email, String organizationUnit, String organization, String city, String state, String country) {
```

impl/src/main/java/io/kroxylicious/testing/kafka/common/CertificateGenerator.java
```diff
@@ -0,0 +1,149 @@
...
+     * This constructor gives the possibility to add a domain as a SAN dns and
+     * an IP address as a SAN IP.
+     *
+     * @param domain the domain
+     * @param ipAddress the ip address
+     */
+    public CertificateGenerator(String domain, String ipAddress) {
+        this.domain = domain;
+        this.ipAddress = ipAddress;
+    }
```
SamBarker: I don't understand why `domain` and `ipAddress` are different to all the other attributes that they should be configured for a whole generator as opposed to. the arguments  of `generateSelfSignedCertificateEntry`.

SamBarker (commented):
I'm sure that the proposed `CertificateGenerator` fulfils the the current needs but I'm not sure it represents the right API.

I think we want to offer an API centred around managing the keystore. 
something like 

```java
import io.netty.pkitesting.CertificateBuilder;
import io.netty.pkitesting.X509Bundle;

public interface KeystoreManager {

    /**
     * Prepares the keystore for use
     */
    void initialiseKeystore();

    /**
     * Creates a CertificateBuilder with the appropriate default values for kroxylicious test usage.
     * @return The partially populated certificate builder.
     */
    CertificateBuilder newCertificateBuilder();

    /**
     * Builds and adds provided certificate builder as a self-signed certificate
     * @param certificateBuilder the builder configuring the certificate
     * @return the self-signed certificate.
     */
    X509Bundle addSelfSignedCertificate(CertificateBuilder certificateBuilder);

    /**
     * Optional we don't need this today!
     * Builds and adds provided certificate builder as a certificate signed by the provided issuer.
     * @param certificateBuilder the builder configuring the certificate
     * @return the signed certificate.
     */
    CertificateBuilder addSignedCertificate(X509Bundle issuer, CertificateBuilder certificateBuilder);
}
```

I think exposing the CertificateBuilder makes things considerably more flexible. What happens in the current proposal if I want to add multiple Subject Alternative Names? What about multiple IP addresses? If we expose the builder those become non issues. 

We could add also add 
```java 

    /**
     * Formats the provided fields into a RFC5280 compliant form
     * @param email
     * @param organizationUnit
     * @param organization
     * @param city
     * @param state
     * @param country
     * @return
     */
    String buildDistinguishedName(String email, String organizationUnit, String organization, String city, String state, String country);
```

As a convince utility method for users (equally we could just make it a string format pattern...

franvila:
> I also note that all file handling has disappeared from this version of the PR. Is that intentional? Without the file writing how do we expect users of the API to consume the generated certificates?

Having the new implementation that provides the `X509Bundle` should be enough to make whatever you need, create a certificate file or use it in the way it is required. But I wonder if we should provide the possibility to generate the file and so on as it was implemented. Maybe it is worth it to make it easier to the final user. 

I'm going to implemented as well.

impl/src/main/java/io/kroxylicious/testing/kafka/common/Keystore.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * New certificate builder.
+     *
+     * @param distinguishedName the distinguished name
+     * @return  the certificate builder
+     */
+    public CertificateBuilder newCertificateBuilder(String distinguishedName) {
+        return newCertificateBuilder().copy()
```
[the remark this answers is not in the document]
franvila: Agreed to avoid this `copy()` but we need to maintain the rest to avoid overriding the original `CertificateBuilder`.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeystoreManager.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /**
+     * Creates a CertificateBuilder with the appropriate default values for kroxylicious test usage.
+     * @param distinguishedName the distinguished name. See {@link KeystoreManager#buildDistinguishedName(String, String, String, String, String, String, String)}
+     *                          for generating a distinguished name
+     * @param sanNames the names for SAN: They can be DNS names and/or IP Addresses
+     * @return  The partially populated certificate builder.
+     */
+    public CertificateBuilder newCertificateBuilder(String distinguishedName, @NonNull List<String> sanNames) {
```
SamBarker: Hmm I understand this is how we want to use it now but somewhat defeats the point of returning the certificate builder. 

I think we merge this signature now to make progress. I'm tempted to say we should mark the the SAN names as optional in javadoc however.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeystoreManager.java
```diff
@@ -0,0 +1,143 @@
...
+        if (this.password == null) {
+            this.password = UUID.randomUUID().toString().replace("-", "");
+        }
+
+        return this.password;
+    }
+
+    /**
+     * Generate certificate file path. It contains both certificates, subject's (key certs) and issuer's (trust CA certs)
+     * See {@link KeystoreManager#getPassword()} for getting the password
```
SamBarker:
```suggestion
    /**
     * Generate a keystore at returned path. It contains both certificates, subject's (key certs) and issuer's (trust CA certs)
     * Use {@link KeystoreManager#getPassword()} for getting the keystore password.
```

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeystoreManager.java
```diff
@@ -0,0 +1,143 @@
...
+     * @return  the path of the generated certificate file
+     * @throws KeyStoreException the key store exception
+     * @throws IOException the io exception
+     * @throws CertificateException the certificate exception
+     * @throws NoSuchAlgorithmException the no such algorithm exception
+     */
+    public Path generateCertificateFile(X509Bundle bundle) throws KeyStoreException, IOException, CertificateException, NoSuchAlgorithmException {
+        KeyStore keyStore = bundle.toKeyStore(getPassword().toCharArray());
+        FileAttribute<Set<PosixFilePermission>> attr = PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwx------"));
+        Path certsDirectory = Files.createTempDirectory("kroxylicious", attr);
```
SamBarker: I'm tempted to make the certs directory a field and allow at least our unit tests to customise it, but I think it might be generally more useful to allow control over where the certs are generated.

SamBarker (commented):
We are getting there!

impl/src/main/java/io/kroxylicious/testing/kafka/common/KeystoreManager.java
SamBarker: Is it worth it? Why not just let users add them directly to the builder?

I'm happy to add the method if there is value I'm just not seeing it right now

SamBarker (approved):
LGTM assuming the build passes.

Thanks for taking the time to get this one right @franvila i think it's needed up in a much better shape

PR #568, opened 2026-01-21
Upgrade to kafka-4.2.0

### LINKED ISSUE #510: Make Kafka Queues testing more convenient
When starting up a single node cluster the default Kafka settings for queue state topics will have a replication factor and min.isr that prevent the topic being created. So users would have to customize the server properties to work with queues.

When queues go GA (undefined, it is currently in Preview as at Kafka 4.1.0) it would be nice to have out-of-the-box support for them so uses who inject the simplest `KafkaCluster` into their tests can use it with share consumers. So we should configure down the defaults for:

* [share.coordinator.state.topic.replication.factor](https://kafka.apache.org/documentation/#brokerconfigs_share.coordinator.state.topic.replication.factor) to 1
* [share.coordinator.state.topic.min.isr](https://kafka.apache.org/documentation/#brokerconfigs_share.coordinator.state.topic.min.isr) to 1

robobario:
### Type of change

- Enhancement / new feature

### Description

Upgrade to Kafka 4.2.0.

Also resolves #510 

### Checklist

- [x] Write tests
- [ ] Update documentation
- [x] Update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Please enumerate all user-facing changes using format `<githib issue/pr number>:
 
 ## SNAPSHOT
 
+* [#568](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/568): Upgrade to kafka-4.2.0
 * [#418](https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/418): Add CloseableShareConsumer (KIP-932)
 
 ## 0.14.0
--- a/impl/src/test/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaClusterTest.java
+++ b/impl/src/test/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaClusterTest.java
@@ -7,7 +7,6 @@ package io.kroxylicious.testing.kafka.testcontainers;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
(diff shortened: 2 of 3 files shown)


robobario:
bleh code cov won't pick up the reflective code that's only there for older brokers

impl/src/main/java/io/kroxylicious/testing/kafka/testcontainers/TestcontainersKafkaCluster.java
```diff
@@ -266,7 +266,7 @@ private KafkaContainer buildKafkaContainer(KafkaClusterConfig.ConfigHolder holde
         if (clusterConfig.isKafkaVersion41OrHigher()) {
             kafkaContainer
                     .withEnv("CLUSTER_ID", holder.kraftClusterId())
-                    .withCopyToContainer(Transferable.of(propertiesToBytes(properties), 0644), "/etc/kafka/docker/server.properties");
+                    .withCopyToContainer(Transferable.of(propertiesToBytes(properties), 0644), "/mnt/shared/config/server.properties");
```
robobario: this is the documented place for config overrides

https://github.com/apache/kafka/tree/4.2/docker/examples#use-input-file-for-providing-configs

k-wall: I see that this has always been the recommended place, so I presume that with will work fine with older Apache Kafka images.

k-wall (approved):
Thanks for picking this up, good to know there are no icebergs.

k-wall:
Intending to ignore the code coverage complaint.. This relates to reflective code which is present to support older versions of Kafka. We know this code is covered by the integration tests.

PR #634, opened 2026-06-19
feat: allow @BrokerConfig to override internal topic replication factors

tombentley:
### Type of change

- Enhancement / new feature

### Description

The hardcoded defaults for internal topic configs (offsets.topic, transaction.state.log, share.coordinator.state.topic replication factors) now use putIfAbsent, so @BrokerConfig annotations can override them. This enables tests that need replicated internal topics, such as coordinator change resilience tests that stop a broker and verify the coordinator re-elects.

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


k-wall (approved):
Code change LGTM.  It would be good to have a supporting test.

PR #649, opened 2026-07-14
Keytool certificate generator

tombentley:
### Type of change

- Bugfix

### Description

Complete the migration from deprecated KeytoolCertificateGenerator to
KeystoreManager, which was introduced in v0.13.0 (commit https://github.com/kroxylicious/kroxylicious-junit5-extension/commit/981eb9c1acc0fe7e83622a143c19f301de656387) but
never fully migrated.

Changes:

**KafkaClusterConfig:**
- Replace KeytoolCertificateGenerator fields with KeystoreManager
- Add brokerKeystorePath, brokerCertBundle, clientKeystorePath, clientCertBundle fields
- Update configureTls() to use KeystoreManager API with Netty's CertificateBuilder
- Update buildSecurityProtocolConfig() to use Path-based keystore references
- Remove processTls() helper method (no longer needed)
- Add imports for io.netty.pkitesting.CertificateBuilder and X509Bundle

**KafkaClusterTest:**
- Replace KeytoolCertificateGenerator fields with KeystoreManager
- Update https://github.com/beforeeach to generate broker certificate using KeystoreManager
- Update createClientCertificate() to use KeystoreManager API
- Replace all builder calls from .brokerKeytoolCertificateGenerator()
  to .keystoreManager(), .brokerKeystorePath(), .clientKeystorePath()
- Add imports for java.nio.file.Path and KeystoreManager

Benefits:
- Uses Netty's CertificateBuilder instead of keytool subprocess
- Cleaner API with X509Bundle intermediate representation
- Single keystore file contains both key and CA certificates
- Eliminates all deprecation warnings (86 occurrences)

The KeystoreManager approach generates PKCS12 keystores that include
both the certificate/key entry and the CA trust entry, simplifying
trust store configuration.

Resolves deprecation warnings:
  [WARNING] KeytoolCertificateGenerator has been deprecated (86 occurrences)

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>
Signed-off-by: Tom Bentley <tbentley@redhat.com>

### DIFF
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -384,9 +384,10 @@ public class KafkaClusterConfig {
                     else {
                         server.put(BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG, "required");
                     }
-                    // The broker must trust the client's CA certificate so that it can authenticate client certificates.
-                    server.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, clientKeystorePath.toAbsolutePath().toString());
-                    server.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, keystoreManager.getPassword(clientKeystorePath));
+                    // Note: With KeystoreManager, the broker's keystore already contains the CA cert for trust
+                    // so we don't need a separate truststore operation like the old generateTrustStore() call
+                    server.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, brokerKeystorePath.toAbsolutePath().toString());
+                    server.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, keystoreManager.getPassword(brokerKeystorePath));
                     server.put(SslConfigs.SSL_TRUSTSTORE_TYPE_CONFIG, "PKCS12");
                 }
             }
@@ -603,14 +604,18 @@ public class KafkaClusterConfig {
         String clientTrustStoreFilePath;
         String clientTrustStorePassword;
         if (clientKeystorePath != null && clientKeystorePath.toFile().exists()) {
-            // SSL client auth: present the client certificate as the client's identity.
+            // SSL client auth case - use client's keystore for both key and trust
             kafkaConfig.put(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, clientKeystorePath.toAbsolutePath().toString());
             kafkaConfig.put(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, keystoreManager.getPassword(clientKeystorePath));
             kafkaConfig.put(SslConfigs.SSL_KEY_PASSWORD_CONFIG, keystoreManager.getPassword(clientKeystorePath));
+            clientTrustStoreFilePath = clientKeystorePath.toAbsolutePath().toString();
+            clientTrustStorePassword = keystoreManager.getPassword(clientKeystorePath);
         }
-        // The client must trust the broker's CA certificate regardless of whether client auth is used.
-        clientTrustStoreFilePath = brokerKeystorePath.toAbsolutePath().toString();
-        clientTrustStorePassword = keystoreManager.getPassword(brokerKeystorePath);
+        else {
+            // No client auth - use broker's keystore for trust only
+            clientTrustStoreFilePath = brokerKeystorePath.toAbsolutePath().toString();
+            clientTrustStorePassword = keystoreManager.getPassword(brokerKeystorePath);
+        }
         kafkaConfig.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, clientTrustStoreFilePath);
         kafkaConfig.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, clientTrustStorePassword);
         kafkaConfig.put(SslConfigs.SSL_TRUSTSTORE_TYPE_CONFIG, "PKCS12");
--- a/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
+++ b/impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
@@ -5,8 +5,13 @@
  */
 package io.kroxylicious.testing.kafka.common;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.lang.annotation.Annotation;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -19,6 +24,7 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -84,8 +81,11 @@ public class KafkaClusterConfig {
     private static final Pattern KAFKA_4_1_OR_HIGHER = Pattern.compile("^4\\.[1-9]\\d*\\..*$");
 
     private TestInfo testInfo;
-    private KeytoolCertificateGenerator brokerKeytoolCertificateGenerator;
-    private KeytoolCertificateGenerator clientKeytoolCertificateGenerator;
+    private KeystoreManager keystoreManager;
+    private Path brokerKeystorePath;
+    private X509Bundle brokerCertBundle;
+    private Path clientKeystorePath;
+    private X509Bundle clientCertBundle;
```
SamBarker: The old API had two fields (`brokerKeytoolCertificateGenerator`, `clientKeytoolCertificateGenerator`) — the new API has five (`keystoreManager`, `brokerKeystorePath`, `brokerCertBundle`, `clientKeystorePath`, `clientCertBundle`).

In the main kroxylicious repo, exactly one place constructs `KafkaClusterConfig` directly with cert fields (`AbstractTlsIT`), and the rest go through `@Tls`. Even `AbstractTlsIT` just passes generators today — it doesn't need fine-grained control over paths and bundles.

The `@Tls` annotation path already gets this right: set `keystoreManager` and let `configureTls()` handle the rest lazily. Could the builder API match that shape? i.e. callers pass a `keystoreManager` (and optionally pre-built bundles for the mTLS case), and `KafkaClusterConfig` owns generating the keystore paths internally.

That would keep `X509Bundle` and `Path` as implementation details rather than builder fields, and eliminate the duplicate cert-generation ceremony in `KafkaClusterTest.before()`.

impl/src/test/java/io/kroxylicious/testing/kafka/KafkaClusterTest.java
```diff
@@ -714,14 +718,23 @@ private void performClusterOperation(Admin admin) {
...
-    private void createClientCertificate() throws GeneralSecurityException, IOException {
-        this.clientKeytoolCertificateGenerator = new KeytoolCertificateGenerator();
-        this.clientKeytoolCertificateGenerator.generateSelfSignedCertificateEntry("clientTest@kroxylicious.io", "client", "Dev", "Kroxylicious.ip", null, null, "US");
+        this.keystoreManager = new KeystoreManager();
+        // Create broker certificate
+        String brokerDn = keystoreManager.buildDistinguishedName("test@kroxylicious.io", "localhost", "Dev",
+                "Kroxylicious.io", null, null, "US");
+        var brokerBuilder = keystoreManager.newCertificateBuilder(brokerDn).addSanDnsName("localhost");
+        var brokerBundle = keystoreManager.createSelfSignedCertificate(brokerBuilder);
+        this.brokerKeystorePath = keystoreManager.generateCertificateFile(brokerBundle);
```
SamBarker: This setup duplicates the cert generation that `configureTls()` already does when `brokerCertBundle == null` (KafkaClusterConfig.java lines ~375–382). The test has to know the internal shape of cert generation to set up the object it's testing — that's a signal the component isn't owning its own initialisation.

If `KafkaClusterConfig` fully owned cert lifecycle (accepting just a `KeystoreManager`), this setup would collapse to `this.keystoreManager = new KeystoreManager()` — matching what the `@Tls` path does.

impl/src/main/java/io/kroxylicious/testing/kafka/common/KafkaClusterConfig.java
```diff
@@ -371,35 +362,42 @@ private boolean isAdditionalNode(int nodeId) {
...
             try {
-                brokerKeytoolCertificateGenerator.generateSelfSignedCertificateEntry("test@kroxylicious.io", listener.advertised().host(), "Dev",
-                        "Kroxylicious.io", null,
-                        null,
-                        "US");
-                if (clientKeytoolCertificateGenerator != null && Path.of(clientKeytoolCertificateGenerator.getCertFilePath()).toFile().exists()) {
+                // Generate broker certificate if not already done
+                if (brokerCertBundle == null) {
+                    String dn = keystoreManager.buildDistinguishedName("test@kroxylicious.io", listener.advertised().host(), "Dev",
+                            "Kroxylicious.io", null, null, "US");
```
SamBarker: Nit: every call site passes `null` for `city` and `state`, which produces a DN containing `L=null, ST=null` (literal string "null"). The old `KeytoolCertificateGenerator.getDomainName` had the same issue, but since `KeystoreManager` is the new canonical API this seems like a good opportunity to fix it — either skip null components in `buildDistinguishedName`, or drop the parameters that no call site uses.

tombentley:
@SamBarker you raise some fair points there. However I wonder where we should be going with all of this:

1. I just prompted Claude to ~"fix the compiler warnings" and that ~"removing `@Deprecated` is not the right fix". I think Claude reasonably concluded that replacing the lingering usages of `KeytoolCertificateGenerator` with `KeystoreManager` was the right way to fix that.
2. But the API of `KeystoreManager` is less than ideal for how we're using it (which is essentially your point here).
3. `KeystoreManager` is a wrapper around the Netty `CertificateBuilder` and `X509Bundle` which we went for as the shortest path to getting rid of the fork/exec pattern we had going on in `KeytoolCertificateGenerator`.
4. In the main repo we also have `io.kroxylicious.testing.certificate.CertificateGenerator` which seems to be doing similarish things, though depending on bouncy castle directly, rather than the Netty `CertificateBuilder` (which itseld depends on BC).
5. It seems to me that it would be better to have a _one_ piece of test support infrastructure for generating certs and keystores, rather than two. That would need:

    * to be in `kroxylicious-junit5-extension`, 
    * to be public API
    * to have a slightly broader API than is needed just for `KafkaClusterConfig`

    I think this is what #131 should really aiming for.

I only came here to fix the compiler warnings left after #544. The real problem was #544 was incomplete and  deprecating `KeytoolCertificateGenerator` didn't really achieve much: It's not in a public API module, and its only usage was in the impl module, where it wasn't replaced by the `KeystoreManager`. 

I've re-opened #131 and I think the tactical fix for the deprecation warnings right now is simply to undeprecate `KeytoolCertificateGenerator`.

tombentley:
@SamBarker PTAL

k-wall:
Was there a Git push error on this branch.  The branch seems to have next to no content.

k-wall:
Ah, I get it now - the outcome is un-deprecate the KeytoolCertificateGenerator and https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/131 reopened.  Maybe closing this PR and a new one with  just a cherry pick of [4774a8b](https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/649/commits/4774a8b2d90a7174d7c5f8a78fe0083e16f83d26) would be least confusing.

If I've groked this right, that makes https://github.com/kroxylicious/kroxylicious/pull/4045 invalid.

tombentley:
> the outcome is un-deprecate the KeytoolCertificateGenerator and https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/131 reopened.

Yes. I mean: I'm willing to be talked down! But right now, I can't tell (from the Issue and PRs I can see) what the overall plan is. The path that #544 _seems to be on_ is to spread usage of `KeytoolCertificateGenerator` (which, I say again, is in the `impl` module, so not intended to be a public API) further into the kroxylicious/kroxylicious repo (cementing it as _de facto_ a public API). Maybe there's thinking I can't see? But if not then I think we should pause and figure out what the overall plan is, taking into account the `io.kroxylicious.testing.certificate.CertificateGenerator` point I made above too, and also Sam's original observation that the new API is not a great replacement. 

> Maybe closing this PR and a new one with just a cherry pick of [4774a8b](https://github.com/kroxylicious/kroxylicious-junit5-extension/pull/649/commits/4774a8b2d90a7174d7c5f8a78fe0083e16f83d26) would be least confusing.

Let's agree on the plan first.

robobario:
Yep agreed that #131 is not done yet and things aren't too clear. I think undeprecating now while we figure it out would be better.  I'm moving my thoughts to that Issue now.

robobario (approved):

PR #652, opened 2026-07-19
Javadoc fixes

tombentley:
### Type of change

- Documentation

### Description

This PR aims to fix the Javadoc warnings which are emitted during the build.

### DIFF
--- a/api/src/main/java/io/kroxylicious/testing/kafka/api/KroxyliciousTestInfo.java
+++ b/api/src/main/java/io/kroxylicious/testing/kafka/api/KroxyliciousTestInfo.java
@@ -11,6 +11,15 @@ import java.util.Set;
 
 import org.junit.jupiter.api.TestInfo;
 
+/**
+ * Implementation of {@link TestInfo} carrying display name, test class, test method and tags
+ * for use within Kroxylicious test infrastructure.
+ *
+ * @param displayName the display name of the test
+ * @param testClass   the test class, if available
+ * @param testMethod  the test method, if available
+ * @param tags        the tags associated with the test
+ */
 public record KroxyliciousTestInfo(String displayName, Optional<Class<?>> testClass, Optional<Method> testMethod, Set<String> tags) implements TestInfo {
     @Override
     public String getDisplayName() {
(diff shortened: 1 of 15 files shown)


tombentley:
@robobario would you be able to take a look at this point, when you've got a moment? Thanks!

robobario (approved):
LGTM, was thinking we should change javadoc warns to a build failure. The only outstanding one is `KafkaClusterConfig` due to the Builder generation creating it's only constructor. Maybe it's worth dropping the javadoc lint for that one file so we can get a build failure on all the others? WDYT

```
--- a/pom.xml
+++ b/pom.xml
@@ -637,7 +637,14 @@
                                 <goals>
                                     <goal>jar</goal>
                                 </goals>
-
+                                <configuration>
+                                    <failOnWarnings>true</failOnWarnings>
+                                    <sourceFileExcludes>
+                                        <!-- @Builder generates an all-args constructor that Javadoc cannot see,
+                                             causing a spurious 'default constructor' missing-comment warning -->
+                                        <sourceFileExclude>**/KafkaClusterConfig.java</sourceFileExclude>
+                                    </sourceFileExcludes>
+                                </configuration>
                             </execution>
                         </executions>
                     </plugin>
```

hmm on second thought, is that excluding it from the javadoc jar too :grimacing: edit yeahp, nevermind. The doclint only lets you exclude checks by package so we're a bit stuffed here.

robobario:
sonar job failed due to what looks like a github API failure

> No server is currently available to service your request. Sorry about that. Please try resubmitting your request and contact us if the problem persists.

kicked it off again

k-wall (approved):
LGTM

tombentley:
> LGTM, was thinking we should change javadoc warns to a build failure. The only outstanding one is KafkaClusterConfig due to the Builder generation creating it's only constructor.

Excluding it from Javadoc is not really the right solution. It should be documented, and we want the documentation to be correct. The problem arises because of the use of Lombok, which adds a constructor without documentation. Claude's already tried a few things to fix it, and failed. The longer term solution would be to get rid of Lombok.

PR #655, opened 2026-07-20
fix(test): re-query cluster metadata in assertClusterIdAndSize

tombentley:
### Type of change

- Bugfix

### Description

The previous implementation captured a single `DescribeClusterResult` before the `await()` loop and then checked `dc.nodes()` repeatedly inside it. Because `dc.nodes()` returns the same already-resolved `KafkaFuture` on every call, the loop never re-fetched cluster state from the broker, so a second broker that hadn't registered at the time of the first call was never seen.

Fix both `assertClusterIdAndSize` and `assertClusterSize` to call `describeCluster()` fresh inside the `untilAsserted` lambda so each poll issues a new metadata request against the live cluster.

Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>


tombentley:
While working on generally tidying up the code I encountered a flaky test. This is the fix.

robobario (approved):
LGTM

PR #673, opened 2026-07-31
refactor: replace JUnit assertion in ScramInitialiser

### LINKED ISSUE #667: Replace JUnit Assertions.assertTrue in ScramInitialiser with plain Java check
## Parent

Part of #665 (Phase 1 — Preparatory).

## Summary

`ScramInitialiser.java:22` uses `org.junit.jupiter.api.Assertions.assertTrue` in **production code**. This should be replaced with a plain Java check (e.g. an `if` + `throw IllegalStateException`).

This is a pre-existing issue independent of the JUnit 6 migration, but fixing it reduces the surface area of JUnit API usage in non-test code.

## Acceptance Criteria

- `Assertions.assertTrue` removed from `ScramInitialiser`
- Replaced with an appropriate plain Java check (e.g. `if (!condition) throw new IllegalStateException(...)`)
- No JUnit imports remain in `ScramInitialiser`
- All existing tests pass

lntutor:
### Type of change

- Refactoring

### Description

Replaces the JUnit `assertTrue` call in `ScramInitialiser` production code with an explicit Java state check and a descriptive `IllegalStateException`.

### Additional Context

This removes the production dependency on JUnit assertions while preserving Awaitility's retry behavior until every configured SCRAM user is visible.

Fixes #667

### Checklist

- [x] Write tests
- [ ] Update documentation (not needed)
- [ ] Update CHANGELOG.md (not needed)

### Verification

- `mvn -pl impl -am -Dtest=ScramInitialiserTest -Dsurefire.failIfNoSpecifiedTests=false process-sources test` on Java 21


lntutor:
Current head `16e9e646ae66485b2a8b9a51adb2744d46b28638` is mergeable, the visible DCO check is green on that exact SHA, and there are no review threads or review-body comments open on the PR at the moment. If this assertion cleanup still matches the intended direction, it looks ready for maintainer review/merge when convenient.

impl/src/test/java/io/kroxylicious/testing/kafka/common/ScramInitialiserTest.java
```diff
@@ -0,0 +1,20 @@
...
+import java.nio.charset.StandardCharsets;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ScramInitialiserTest {
+
+    @Test
+    void shouldNotReferenceJUnitAssertions() throws IOException {
```
robobario: We should test the IllegalStateException logic instead. Lets test the failure mode, not some quality of the bytecode. This looks brittle to me.

robobario (changes requested):
Thanks @lntutor! The implementation looks good to me, but I'd rather merge with no unit test than the bytecode string comparison.

impl/src/test/java/io/kroxylicious/testing/kafka/common/ScramInitialiserTest.java
lntutor: Addressed in 5f247cdd4bbf3e9532bf62fc535d13581f4f0718 by removing the brittle bytecode-string test as requested. Verification on this exact SHA: all 349 affected-module tests pass, formatting and Checkstyle pass, and the license check reports 62/62 files valid. The DCO check is also green.

robobario (approved):
LGTM

